PR #12, opened 2022-10-06
Refactor InFlight State

EasterTheBunny:
In processing incoming logs, delete items in in-flight state.

When checking upkeeps, return state as in-flight if the key exists.

The registry should return the block number returned from the contract instead of the block number used to check the upkeep. This keeps the cache keys and perform data wrapper in sync.

### DIFF
--- a/internal/keepers/ocr.go
+++ b/internal/keepers/ocr.go
@@ -25,7 +25,7 @@ func (k *keepers) Observation(ctx context.Context, _ types.ReportTimestamp, _ ty
 
 	ob := observationMessageProto{
 		RandomValue: k.rSrc.Int63(),
-		Keys:        keyList(filterUpkeeps(results, ktypes.Perform)),
+		Keys:        keyList(filterUpkeeps(results, ktypes.Eligible)),
 	}
 
 	b, err := encode(ob)
@@ -73,7 +73,7 @@ func (k *keepers) Report(ctx context.Context, _ types.ReportTimestamp, _ types.Q
 			return false, nil, fmt.Errorf("%w: failed to check upkeep from attributed observation", err)
 		}
 
-		if upkeep.State == ktypes.Perform {
+		if upkeep.State == ktypes.Eligible {
 			// only build a report from a single upkeep for now
 			k.logger.Printf("reporting %s to be performed", upkeep.Key)
 			toPerform = append(toPerform, upkeep)
@@ -91,9 +91,10 @@ func (k *keepers) Report(ctx context.Context, _ types.ReportTimestamp, _ types.Q
 		return false, nil, fmt.Errorf("%w: failed to encode OCR report", err)
 	}
 
+	// TODO: move this to ShouldAcceptFinalizedReport
 	// update internal state of upkeeps to ensure they aren't reported or observed again
 	for i := 0; i < len(toPerform); i++ {
-		if err := k.service.SetUpkeepState(ctx, toPerform[i].Key, ktypes.Reported); err != nil {
+		if err := k.service.SetUpkeepState(ctx, toPerform[i].Key, ktypes.InFlight); err != nil {
 			return false, nil, fmt.Errorf("%w: failed to update internal state while generating report", err)
 		}
 	}
@@ -105,6 +106,7 @@ func (k *keepers) Report(ctx context.Context, _ types.ReportTimestamp, _ types.Q
 // from OCR2. The implementation is the most basic possible in that it only
 // checks that a report has data to send.
 func (k *keepers) ShouldAcceptFinalizedReport(_ context.Context, _ types.ReportTimestamp, r types.Report) (bool, error) {
+	// TODO: decode report, set reported status for each upkeep
 	// TODO: isStale check on last performed block number
 	shouldAccept := len(r) != 0
 	if shouldAccept {
--- a/internal/keepers/service.go
+++ b/internal/keepers/service.go
@@ -19,7 +19,7 @@ type onDemandUpkeepService struct {
 	perfLogs       types.PerformLogProvider
 	shuffler       shuffler[types.UpkeepKey]
 	cache          *cache[types.UpkeepResult]
-	stateCache     *cache[types.UpkeepState]
+	inFlightCache  *cache[types.UpkeepKey]
 	cacheCleaner   *intervalCacheCleaner[types.UpkeepResult]
 	workers        *workerGroup[types.UpkeepResult]
 	stopProcs      chan struct{}
@@ -33,15 +33,15 @@ type onDemandUpkeepService struct {
 // from this service.
 func newOnDemandUpkeepService(ratio sampleRatio, registry types.Registry, perfLogs types.PerformLogProvider, logger *log.Logger, cacheExpire time.Duration, staleWindow time.Duration, cacheClean time.Duration, workers int, workerQueueLength int) *onDemandUpkeepService {
 	s := &onDemandUpkeepService{
-		logger:     logger,
-		ratio:      ratio,
-		registry:   registry,
-		perfLogs:   perfLogs,
-		shuffler:   new(cryptoShuffler[types.UpkeepKey]),
-		cache:      newCache[types.UpkeepResult](cacheExpire),
-		stateCache: newCache[types.UpkeepState](staleWindow),
-		workers:    newWorkerGroup[types.UpkeepResult](workers, workerQueueLength),
-		stopProcs:  make(chan struct{}),
+		logger:        logger,
+		ratio:         ratio,
+		registry:      registry,
+		perfLogs:      perfLogs,
+		shuffler:      new(cryptoShuffler[types.UpkeepKey]),
+		cache:         newCache[types.UpkeepResult](cacheExpire),
+		inFlightCache: newCache[types.UpkeepKey](staleWindow),
+		workers:       newWorkerGroup[types.UpkeepResult](workers, workerQueueLength),
+		stopProcs:     make(chan struct{}),
 	}
 
 	cl := &intervalCacheCleaner[types.UpkeepResult]{
@@ -89,12 +89,18 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
 	if err != nil {
 		return result, err
 	}
-	state, stateCached := s.stateCache.Get(string(id))
 
+	// if an upkeep is in-flight (reported but not confirmed) all checks should
+	// return the InFlight status
+	_, isInFlight := s.inFlightCache.Get(string(id))
+
+	// the cache is a collection of keys (block & id) that map to cached
+	// results. if the same upkeep is checked at a block that has already been
+	// checked, return the cached result
 	result, cached := s.cache.Get(string(key))
 	if cached {
-		if stateCached {
-			result.State = state
+		if isInFlight {
+			result.State = types.InFlight
 		}
 		return result, nil
 	}
@@ -108,16 +114,18 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
 
 	result = types.UpkeepResult{
 		Key:   key,
-		State: types.Skip,
+		State: types.NotEligible,
 	}
 
 	if needsPerform {
-		result.State = types.Perform
+		result.State = types.Eligible
 		result.PerformData = u.PerformData
 	}
 
-	if stateCached {
-		result.State = state
+	// the upkeep should be in the cache, but in the case it's not and it is
+	// in-flight, set the state appropriately
+	if isInFlight {
+		result.State = types.InFlight
 	}
 
 	s.cache.Set(string(key), result, defaultExpiration)
@@ -135,8 +143,9 @@ func (s *onDemandUpkeepService) SetUpkeepState(ctx context.Context, uk types.Upk
 
 	// set the state cache to the default expiration using the upkeep identifer
 	// as the key
-	s.stateCache.Set(string(id), state, defaultExpiration)
+	s.inFlightCache.Set(string(id), uk, defaultExpiration)
 
+	// if the result for this key is cached, update the state to in-flight
 	result, cached := s.cache.Get(string(uk))
 	if !cached {
 		// if the value is not in the cache, do a hard check
@@ -161,9 +170,28 @@ func (s *onDemandUpkeepService) start() {
 			for {
 				select {
 				case log := <-ch:
-					// TODO: add delete method to cache; hack sets a new status and 1
-					// nanosecond expire time
-					s.stateCache.Set(string(log.Key), types.Performed, 1)
+					// TODO: check perform log confirmations; expect the log
+					// provider to send a log again at a higher block number
+					// which will have more confirmations
+
+					// update the cache with a performed state. in the case
+					// that this key is queried again, it will not be added to
+					// a report
+					r, cached := s.cache.Get(string(log.Key))
+					if cached {
+						r.State = types.Performed
+						s.cache.Set(string(log.Key), r, defaultExpiration)
+					}
+
+					id, err := s.registry.IdentifierFromKey(log.Key)
+					if err != nil {
+						s.logger.Printf("failed to get id from upkeep key: %s", err)
+						continue
+					}
+
+					// remove the in-flight key since this id has been confirmed
+					s.inFlightCache.Delete(string(id))
+
 				case <-s.stopProcs:
 					// closing the channel indicates to the sender that the channel is
 					// no longer being used. This will cause the sending service to
@@ -192,7 +220,6 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
 	samples := newSyncedArray[*types.UpkeepResult]()
 
 	var cacheHits int
-
 	var wg sync.WaitGroup
 
 	// start the results listener
@@ -215,7 +242,7 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
 					success++
 					// cache results
 					s.cache.Set(string(result.Data.Key), result.Data, defaultExpiration)
-					if result.Data.State == types.Perform {
+					if result.Data.State == types.Eligible {
 						samples.Append(&result.Data)
 					}
 				} else {
@@ -238,22 +265,28 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
 		if err != nil {
 			return samples.Values(), err
 		}
-		state, stateCached := s.stateCache.Get(string(id))
-		if stateCached && state == types.Reported {
+
+		// the reported state is in the state cache for the upkeep id
+		// if the upkeep id has been reported, we consider this to be 'in-flight'
+		// and should not be included in samples
+		_, isInFlight := s.inFlightCache.Get(string(id))
+		if isInFlight {
 			cacheHits++
 			continue
 		}
 
-		// skip if reported
+		// no RPC lookups need to be done if a result has already been cached
 		result, cached := s.cache.Get(string(key))
 		if cached {
 			cacheHits++
-			if result.State == types.Perform {
+			if result.State == types.Eligible {
 				samples.Append(&result)
 			}
 			continue
 		}
 
+		// for every job added to the worker queue, add to the wait group
+		// all jobs should complete before completing the function
 		wg.Add(1)
 		if err := s.workers.Do(ctx, makeWorkerFunc(s.logger, s.registry, key, ctx)); err != nil {
 			if errors.Is(err, ErrContextCancelled) {
@@ -323,6 +356,9 @@ func makeWorkerFunc(logger *log.Logger, registry types.Registry, key types.Upkee
 
 		// close go-routine to prevent memory leaks
 		done <- struct{}{}
+
+		// err could exist if header block and check block don't match
+		// TODO: maybe this should be a soft error
 
 		return u, err
 	}
--- a/pkg/chain/evmregistry.go
+++ b/pkg/chain/evmregistry.go
@@ -18,10 +18,11 @@ const ActiveUpkeepIDBatchSize int64 = 10000
 const separator string = "|"
 
 var (
-	ErrRegistryCallFailure   = fmt.Errorf("registry chain call failure")
-	ErrBlockKeyNotParsable   = fmt.Errorf("block identifier not parsable")
-	ErrUpkeepKeyNotParsable  = fmt.Errorf("upkeep key not parsable")
-	ErrInitializationFailure = fmt.Errorf("failed to initialize registry")
+	ErrRegistryCallFailure      = fmt.Errorf("registry chain call failure")
+	ErrBlockKeyNotParsable      = fmt.Errorf("block identifier not parsable")
+	ErrUpkeepKeyNotParsable     = fmt.Errorf("upkeep key not parsable")
+	ErrInitializationFailure    = fmt.Errorf("failed to initialize registry")
+	ErrBlockAtCheckDoesNotMatch = fmt.Errorf("header block and check block do not match")
 )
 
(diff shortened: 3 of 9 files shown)


internal/keepers/service.go
```diff
@@ -89,12 +89,18 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
 	if err != nil {
 		return result, err
 	}
-	state, stateCached := s.stateCache.Get(string(id))
 
+	// if an upkeep is in-flight (reported but not confirmed) all checks should
+	// return the InFlight status
+	_, isInFlight := s.inFlightCache.Get(string(id))
```
infiloop2, line 95: wdyt about moving this logic outside checkUpkeep (to the plugin)? 

Would limit the purpose of this function to checkUpkeep using checkCache, while inFlightCache remains a plugin concept

internal/keepers/service.go
```diff
@@ -19,7 +19,7 @@ type onDemandUpkeepService struct {
 	perfLogs       types.PerformLogProvider
 	shuffler       shuffler[types.UpkeepKey]
 	cache          *cache[types.UpkeepResult]
```
infiloop2, line 21: nit. rename to checkCache / checkUpkeepCache or something?

internal/keepers/service.go
```diff
@@ -89,12 +89,18 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
...
+
+	// the cache is a collection of keys (block & id) that map to cached
+	// results. if the same upkeep is checked at a block that has already been
+	// checked, return the cached result
 	result, cached := s.cache.Get(string(key))
 	if cached {
-		if stateCached {
-			result.State = state
+		if isInFlight {
+			result.State = types.InFlight
```
infiloop2, line 103: I think we can separate out these states.

1. State for checkUpkeep - eligible or not -> depends on geth call
2. State for inFlight - inFlight or not -> depends on plugin execution

internal/keepers/service.go
```diff
@@ -135,8 +143,9 @@ func (s *onDemandUpkeepService) SetUpkeepState(ctx context.Context, uk types.Upk
 
 	// set the state cache to the default expiration using the upkeep identifer
 	// as the key
-	s.stateCache.Set(string(id), state, defaultExpiration)
+	s.inFlightCache.Set(string(id), uk, defaultExpiration)
```
infiloop2, line 146: not sure why we need to change inFlight cache here. Should only be changed in shouldAcceptReport and upon receiving a log

internal/keepers/service.go
```diff
@@ -161,9 +170,28 @@ func (s *onDemandUpkeepService) start() {
...
+					// which will have more confirmations
+
+					// update the cache with a performed state. in the case
+					// that this key is queried again, it will not be added to
+					// a report
+					r, cached := s.cache.Get(string(log.Key))
+					if cached {
+						r.State = types.Performed
+						s.cache.Set(string(log.Key), r, defaultExpiration)
+					}
```
infiloop2, line 184: not sure why we need to update both cache and inFlightCache here. Can separate out the concerns

infiloop2 (changes requested):
Added comments on how we can completely isolate the two caches which have very different purposes

1. CheckCache -> Used to cache rpc call results. Key is upkeepID + block Number. Value contains the whole check result. Used only in evmregistry, plugin doesn't need to know about it.

2. InFlightCache -> Used to not double report on upkeeps. Only a plugin concept, evmregistry doesn't need to know about it

pkg/types/types.go
```diff
@@ -52,10 +54,9 @@ type UpkeepResult struct {
...
 
 const (
-	Eligible UpkeepState = iota
-	Skip
-	Perform
-	Reported
+	NotEligible UpkeepState = iota
+	Eligible
+	InFlight
 	Performed
```
infiloop2, line 60: Not sure if we really need a performed state

pkg/types/types.go
```diff
@@ -52,10 +54,9 @@ type UpkeepResult struct {
 type UpkeepState uint
```
infiloop2, line 54: I would suggest to split this into two different states.

One is a state per upkeep ID (inFlight or not)
One is a state per key (upkeepID+block) (eligible or not)

internal/keepers/service.go
```diff
@@ -161,9 +170,28 @@ func (s *onDemandUpkeepService) start() {
 			for {
 				select {
 				case log := <-ch:
-					// TODO: add delete method to cache; hack sets a new status and 1
-					// nanosecond expire time
-					s.stateCache.Set(string(log.Key), types.Performed, 1)
+					// TODO: check perform log confirmations; expect the log
+					// provider to send a log again at a higher block number
+					// which will have more confirmations
```
infiloop2, line 175: can we move this logic outside this block so that we only get logs post confirmation? I would expect it to be a config sent to log poller upon initialisation ?

internal/keepers/service.go
EasterTheBunny: The problem I see with this is that the service needs to know which upkeeps to return a cached result for. Maybe inflight upkeeps could be passed to `SampleUpkeeps`? That would separate concerns a bit. wdyt?

internal/keepers/service.go
EasterTheBunny: Makes sense. Originally it was the only cache, so naming it something that says what it's doing will help with readability.

internal/keepers/service.go
EasterTheBunny: If we move the inflight ids to a separate data structure and pass that into the service, that would simplify the service a good bit.

internal/keepers/service.go
EasterTheBunny: The method `SetUpkeepState` is called from `Report` at the moment, but will be called from `ShouldAcceptReport` once the report decoding is done.

If we refactor such that we make an inflight data structure separate from the check upkeep service, it becomes a dependency of the OCR concept and passed into necessary functions of the service like `SampleUpkeeps`.

internal/keepers/service.go
EasterTheBunny: The concept is to block out an upkeep id from being checked. Since the status is isolated to the upkeep result, wherever it is passed, the status should indicate that it should not be eligible. Maybe that's just two statuses then? Eligible and NotEligible?

internal/keepers/service.go
EasterTheBunny: That depends on where we put the confirmations config param. If we use the offChainConfig, that config is only available after plugin initialization which would isolate its value to the plugin. If we use the node config, we can apply the confirmation filter to the logs before they reach the plugin.

internal/keepers/service.go
infiloop2: yep I think that would make sense. Let's discuss more offline

internal/keepers/service.go
infiloop2: ccip also uses this as offchain config we can look at their solution https://github.com/smartcontractkit/chainlink-ccip/blob/ccip-develop/core/services/ocr2/plugins/ccip/offchain_config.go#L7-L10

internal/keepers/ocr.go
```diff
@@ -91,48 +76,67 @@ func (k *keepers) Report(ctx context.Context, _ types.ReportTimestamp, _ types.Q
...
+
+	for _, id := range ids {
+		locked, err := k.service.IsUpkeepLocked(ctx, id)
+		if err != nil {
+			return false, fmt.Errorf("failed to check lock for upkeep '%s' in epoch %d for round %d: %s", string(id), rt.Epoch, rt.Round, err)
+		}
+
+		// multiple keys can be in a single report. if one fails to run in the
+		// contract, but others are successful, don't try to transmit again
+		if locked {
```
infiloop2, line 133: nit. Add TODO to rethink this logic when we add batching

internal/keepers/service.go
```diff
@@ -108,82 +113,88 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
...
 
+func (s *onDemandUpkeepService) IsUpkeepLocked(ctx context.Context, id types.UpkeepIdentifier) (bool, error) {
+	_, ok := s.lockoutCache.Get(string(id))
+	return ok, nil
+}
+
 func (s *onDemandUpkeepService) start() {
+	// TODO: if this process panics, restart it
 
 	if s.perfLogs != nil {
```
infiloop2, line 150: nit. Log critical if this is nil

internal/keepers/service.go
```diff
@@ -108,82 +113,88 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
...
-		go func(ch chan types.PerformLog) {
 			for {
+				ctx, cancel := context.WithTimeout(context.Background(), cadence)
 				select {
-				case log := <-ch:
-					// TODO: add delete method to cache; hack sets a new status and 1
-					// nanosecond expire time
-					s.stateCache.Set(string(log.Key), types.Performed, 1)
+				case <-t.C:
+					lastIndex, logs, err := s.perfLogs.ConfirmedPerformsFromIndex(ctx, idx)
```
infiloop2, line 160: todo: confirm re-org behaviour here

infiloop2: we probably want to have a lookback here from idx. something like `idx-1`

internal/keepers/ocr.go
```diff
@@ -91,48 +76,67 @@ func (k *keepers) Report(ctx context.Context, _ types.ReportTimestamp, _ types.Q
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get ids from report: %w", err)
+	}
+
+	if len(ids) == 0 {
+		return false, fmt.Errorf("no ids in report in epoch %d for round %d", rt.Epoch, rt.Round)
+	}
+
+	for _, id := range ids {
+		locked, err := k.service.IsUpkeepLocked(ctx, id)
```
infiloop2, line 126: needs to be on upkeepID, block number key

internal/keepers/service.go
```diff
@@ -108,82 +113,88 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
...
+					if err != nil {
+						s.logger.Printf("failed to get performed logs: %s", err)
+					}
+
+					for _, log := range logs {
+						// remove the item from the cache to clear the state
+						_, cached := s.cache.Get(string(log.Key))
+						if cached {
+							s.cache.Delete(string(log.Key))
+						}
```
infiloop2, line 170: do we really need this

internal/keepers/service.go
```diff
@@ -108,82 +113,88 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, key types.Upkee
...
-		s.subscriptionId, chPerfLogs = s.perfLogs.Subscribe()
+		go func() {
+			cadence := time.Second
+			t := time.NewTimer(cadence)
+			var idx int
 
-		go func(ch chan types.PerformLog) {
 			for {
+				ctx, cancel := context.WithTimeout(context.Background(), cadence)
 				select {
```
infiloop2, line 158: add context cancellation here

pkg/chain/evmreports.go
```diff
@@ -75,6 +69,69 @@ func (b *evmReportEncoder) EncodeReport(toReport []ktypes.UpkeepResult) ([]byte,
...
+
+	reportArgs := abi.Arguments{
+		{Name: "fastGasWei", Type: Uint256},
+		{Name: "linkNative", Type: Uint256},
+		{Name: "upkeepIds", Type: Uint256Arr},
+		{Name: "wrappedPerformDatas", Type: PerformDataArr},
+	}
+
+	m := make(map[string]interface{})
+	err := reportArgs.UnpackIntoMap(m, report)
```
infiloop2, line 86: can use the same way for checkUpkeep()

internal/keepers/factory.go
```diff
@@ -65,21 +66,40 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 	// number needs to either come from a config, or be calculated on actual
 	// performance of the nodes in real time. that is, start at 1 and increment
 	// after some blocks pass until a stable number is reached.
```
infiloop2, line 68: nit. TODO can be removed

internal/keepers/filter.go
```diff
@@ -0,0 +1,176 @@
...
+			logLookup := make(map[string]types.PerformLog)
+			var minConfirmations int64 = 20
+			for _, log := range logs {
+				logLookup[string(log.Key)] = log
+				if _, ok := rc.activeKeys.Get(string(log.Key)); ok {
+					if log.Confirmations < minConfirmations {
+						// set the active key value to false to indicate that
+						// we now have a transaction identified and no new
+						// attempts at a transaction should be made
+						rc.activeKeys.Set(string(log.Key), true, defaultExpiration)
```
infiloop2, line 133: not sure i completely follow, let's discuss

internal/keepers/filter.go
```diff
@@ -0,0 +1,176 @@
...
+					continue
+				}
+
+				rc.idBlocks.Delete(string(id))
+			}
+
+			// perform logs can potentially come and go. for all keys, if a
+			// perform log is not detected, set the active keys value to true to
+			// indicate that we are waiting for a transaction log and any node
+			// should attempt to transmit.
```
infiloop2, line 153: nit. true/false is opposite as of comment.

also not sure if it should be "any node should not attempt to transmit."?

internal/keepers/filter.go
```diff
@@ -0,0 +1,176 @@
...
+			// in the case of reorgs causing a perform to be dropped from the
+			// chain, total confirmations are checked to allow other nodes
+			// to retry a transaction based on the OCR transmit cadence.
+			logLookup := make(map[string]types.PerformLog)
+			var minConfirmations int64 = 20
+			for _, log := range logs {
+				logLookup[string(log.Key)] = log
+				if _, ok := rc.activeKeys.Get(string(log.Key)); ok {
+					if log.Confirmations < minConfirmations {
+						// set the active key value to false to indicate that
```
infiloop2, line 130: nit. true/false is opposite as comment and code

internal/keepers/ocr.go
```diff
@@ -18,17 +18,14 @@ func (k *keepers) Query(_ context.Context, _ types.ReportTimestamp) (types.Query
...
+	results, err := k.service.SampleUpkeeps(ctx, k.filter.Filter())
 	if err != nil {
 		return types.Observation{}, err
 	}
 
-	ob := observationMessageProto{
-		RandomValue: k.rSrc.Int63(),
-		Keys:        keyList(filterUpkeeps(results, ktypes.Perform)),
-	}
+	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
```
infiloop2, line 26: nit. we have two filters (one for in-flight upkeeps and one filter for eligible). Can make the distinction clear in naming

internal/keepers/ocr.go
```diff
@@ -43,40 +40,30 @@ func (k *keepers) Observation(ctx context.Context, _ types.ReportTimestamp, _ ty
...
 		if err != nil {
 			return false, nil, fmt.Errorf("%w: failed to check upkeep from attributed observation", err)
 		}
 
-		if upkeep.State == ktypes.Perform {
+		if upkeep.State == ktypes.Eligible {
 			// only build a report from a single upkeep for now
 			k.logger.Printf("reporting %s to be performed", upkeep.Key)
 			toPerform = append(toPerform, upkeep)
+			err = k.filter.Add(upkeep.Key)
```
infiloop2, line 63: I don't think we should add here. Should be done once a report with signatures is generated

internal/keepers/service.go
```diff
@@ -237,27 +154,30 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
...
 			continue
 		}
 
-		// skip if reported
+		// no RPC lookups need to be done if a result has already been cached
 		result, cached := s.cache.Get(string(key))
 		if cached {
 			cacheHits++
-			if result.State == types.Perform {
+			if result.State == types.Eligible {
```
infiloop2, line 173: not sure why we need this special logic if cached. this should be applicable in either case?

internal/keepers/service.go
```diff
@@ -219,7 +136,7 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
 					success++
 					// cache results
 					s.cache.Set(string(result.Data.Key), result.Data, defaultExpiration)
```
infiloop2, line 138: not sure if easy, but would be great if we can just reuse the checkUpkeep function and not worry about cache in this function

pkg/chain/evmregistry.go
```diff
@@ -168,6 +170,14 @@ func (r *evmRegistryv2_0) CheckUpkeep(ctx context.Context, key types.UpkeepKey)
 	result.CheckBlockHash = ret0.Result.CheckBlockhash
 	result.PerformData = ret0.Result.PerformData
 
+	// reset key to check block number returned from contract
+	// this is the result that will be used in the cache both to reduce repeated
+	// check attempts and provide a lookup for in-flight performs
+	// the reason for this is in the case that the block the upkeep is checked
+	// at is different than what the contract returned. the contract value
+	// takes priority.
+	result.Key = BlockAndIdToKey(big.NewInt(int64(result.CheckBlockNumber+1)), upkeepId)
```
infiloop2, line 179: not sure why you need blockNumber + 1?

pkg/chain/evmregistry.go
```diff
@@ -18,10 +18,11 @@ const ActiveUpkeepIDBatchSize int64 = 10000
...
 var (
-	ErrRegistryCallFailure   = fmt.Errorf("registry chain call failure")
-	ErrBlockKeyNotParsable   = fmt.Errorf("block identifier not parsable")
-	ErrUpkeepKeyNotParsable  = fmt.Errorf("upkeep key not parsable")
-	ErrInitializationFailure = fmt.Errorf("failed to initialize registry")
+	ErrRegistryCallFailure      = fmt.Errorf("registry chain call failure")
+	ErrBlockKeyNotParsable      = fmt.Errorf("block identifier not parsable")
+	ErrUpkeepKeyNotParsable     = fmt.Errorf("upkeep key not parsable")
+	ErrInitializationFailure    = fmt.Errorf("failed to initialize registry")
+	ErrBlockAtCheckDoesNotMatch = fmt.Errorf("header block and check block do not match")
```
infiloop2, line 25: nit. doesn't seem to be used

pkg/chain/evmreports.go
```diff
@@ -75,6 +69,86 @@ func (b *evmReportEncoder) EncodeReport(toReport []ktypes.UpkeepResult) ([]byte,
...
+	}
+
+	if link, ok = m[mKeys[1]].(*big.Int); !ok {
+		return res, fmt.Errorf("link native as wrong type")
+	}
+
+	res = make([]ktypes.UpkeepResult, len(upkeepIds))
+	for i := 0; i < len(upkeepIds); i++ {
+		res[i] = ktypes.UpkeepResult{
+			Key:              BlockAndIdToKey(big.NewInt(int64(performs[i].CheckBlockNumber)), upkeepIds[i]),
```
infiloop2, line 137: needs to be same as one in checkUpkeep

internal/keepers/filter.go
```diff
@@ -0,0 +1,176 @@
...
+	cadence := time.Second
+	timer := time.NewTimer(cadence)
+
+	for {
+
+		select {
+		case <-timer.C:
+			startTime := time.Now()
+			keys := rc.activeKeys.Keys()
+			logs, _ := rc.logs.PerformLogs(context.Background())
```
infiloop2, line 114: note: this needs to look back some range of blocks (i think depends on performLockoutWindow and deltaStage * n). Would be good to have that logic here

internal/keepers/factory.go
```diff
@@ -65,21 +66,40 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
...
+	}
+
+	p, err = strconv.ParseFloat(offChainCfg.TargetProbability, 32)
+	if err != nil {
+		return nil, info, fmt.Errorf("%w: failed to parse configured probability", err)
+	}
+
+	if offChainCfg.TargetInRounds <= 0 {
+		offChainCfg.TargetInRounds = 1
+	}
```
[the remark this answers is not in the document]
EasterTheBunny: The configuration number is based on this equation that calculates probability over multiple rounds. We can leave it at 1, but it is part of the probability sampling equation.

internal/keepers/factory.go
```diff
@@ -65,21 +66,40 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 	// number needs to either come from a config, or be calculated on actual
 	// performance of the nodes in real time. that is, start at 1 and increment
 	// after some blocks pass until a stable number is reached.
-	sample, err := sampleFromProbability(1, c.N-c.F, 0.99999)
+	var p float64
+	if len(offChainCfg.TargetProbability) == 0 {
```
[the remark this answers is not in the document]
EasterTheBunny: That's probably good in the case of misconfigured production instances, but makes development difficult. Defaults define a baseline for run parameters such that the program can spin up and run without a user knowing much about it. I think this should be a discussion for after the completion of the beta, though. It's kinda philosophical. lol

internal/keepers/factory.go
EasterTheBunny: I like this thought, though, on eliminating the need for the second config param. The concept is to achieve the targeted probability while adjusting the number of rounds necessary to achieve it. It's a real-time optimization of upkeep load distribution.

internal/keepers/filter.go
EasterTheBunny: Damn. yes it is. I switched it and changed the function name because the former was confusing.

internal/keepers/filter.go
EasterTheBunny: yup

internal/keepers/ocr.go
EasterTheBunny: The reason is because query/observer/report is a different process than accept/transmit. The former has the ability to continue moving on at its own pace meaning round 2 of the former can begin before round 1 of the latter has completed. We need to be able to move through ids quickly in the first stage because we're doing them one report at a time.

internal/keepers/service.go
EasterTheBunny: All results should be cached, eligible or not.

internal/keepers/service.go
EasterTheBunny: So the workers just call the service's CheckUpkeep? Makes sense. I really want to clean that section up anyway. It'll be a good todo item for later.

pkg/chain/evmregistry.go
EasterTheBunny: This is the part we should talk about over a call. There were some changes to the registry and I'm trying to ensure keys line up everywhere. Rebuilding the key may not be necessary, but let's talk through it to be sure.

pkg/chain/evmreports.go
EasterTheBunny: That's the tricky bit. Is it plus 1 here too? Or not in either place?

internal/keepers/filter.go
EasterTheBunny: Let's also bring that up in discussion. I'm guessing the lookback is chain specific but can it be an implementation detail of the PerformLogsProvider?

internal/keepers/factory.go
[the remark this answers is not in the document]
infiloop2: let's combine all default values here https://github.com/smartcontractkit/ocr2keepers/blob/main/pkg/types/types.go#L79-L81

infiloop2 (dismissed):

PR #22, opened 2022-10-14
Only process unconfirmed logs

infiloop2:


internal/keepers/filter.go
```diff
@@ -131,14 +131,18 @@ func (rc *reportCoordinator) run() {
...
+				confirmed, ok := rc.activeKeys.Get(string(log.Key))
+				if ok && !confirmed {
+					// if we detect a log, remove it from the observation filters
+					// to allow it to be reported on again
+					rc.idBlocks.Delete(string(id))
+
+					// set state of key to indicate that the report was transmitted
+					// setting a key in this way also blocks it in Accept even if
+					// Accept was never called for on a single node for this key
+					rc.activeKeys.Set(string(log.Key), true, defaultExpiration)
```
EasterTheBunny, line 144: I think setting the active key confirmation status should be an idempotent operation for all logs. Maybe set the key after the confirmation status is checked for the idBlock deletion?

I'm going to still accept the review though and make a new tag because if it's working for you, then we should move it forward and possibly address this detail later.

EasterTheBunny (approved):

PR #30, opened 2022-10-18
Limit keys returned by observation programatically

EasterTheBunny:
Max observation limit is total bytes and the key list returned by the observation step should be under that size after encoding. This change programatically builds a sorted key list that fits in within the limit.


internal/keepers/ocr.go
```diff
@@ -52,15 +52,24 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
...
 	// limit
-	if len(keys) > 10 {
-		keys = keys[:10]
+	var tot int
+	var idx int
+	// if the total plus padding for json encoding is less than the max, another
+	// key can be included
+	for tot+(5*(idx+1)) < maxObservationLength && idx < len(keys) {
+		tot += len(keys[idx])
+		idx++
```
infiloop2, line 68: I don't fully understand the logic (especially the magical number 5). 
imo a constant limit is much simpler and the dynamic limit doesn't give much advantage, but will leave it upto you

internal/keepers/service_test.go
```diff
@@ -153,7 +153,7 @@ func TestOnDemandUpkeepService(t *testing.T) {
 		}
 
 		assert.Less(t, len(result), 25)
-		assert.Greater(t, len(result), 5)
+		assert.Greater(t, len(result), 4)
```
infiloop2, line 156: not sure why this required changing?

internal/keepers/ocr.go
EasterTheBunny: I'm not usually keen on magic number either, but it's based on the number of extra bytes for each key that would be added in the encode process.

internal/keepers/util.go
```diff
@@ -55,6 +55,8 @@ func keyList(upkeeps []*ktypes.UpkeepResult) []ktypes.UpkeepKey {
 		ret[i] = up.Key
 	}
 
+	sort.Sort(sortUpkeepKeys(ret))
```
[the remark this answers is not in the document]
EasterTheBunny: Maybe not, but it's safer since we don't have much time to test.

internal/keepers/service_test.go
EasterTheBunny: It's a flaky test that needs to be changed.

PR #32, opened 2022-10-19
Modify observation context before sampler

EasterTheBunny:
The observation context can cancel potentially causing all results from sampling to be lost for that node in that round. Modifying the parent context to reduce the deadline allows a window for all workers to shut down and report as well as enough time for the observation to complete in time with at least some results.

This allows an observation to not be a total loss for that round if RPC calls are taking longer than expected.


infiloop2 (approved):

internal/keepers/ocr.go
```diff
@@ -46,6 +46,19 @@ func (k *keepers) Query(_ context.Context, _ types.ReportTimestamp) (types.Query
...
+		// this forces all workers to close and report early, allowing the
+		// observation round to not be a total loss.
+		deadline.Add(-1 * 100 * time.Millisecond)
+		ctx, cancel = context.WithDeadline(ctx, deadline)
+	}
+
+	defer cancel()
 
 	results, err := k.service.SampleUpkeeps(ctx, k.filter.Filter())
 	if err != nil {
```
infiloop2, line 64: do we need to change this error handling?

PR #40, opened 2022-11-23
Running sampling upkeeps in parallel

begmaroman:


internal/keepers/service.go
```diff
@@ -9,36 +9,51 @@ import (
 	"sync"
 	"time"
 
+	ethtypes "github.com/ethereum/go-ethereum/core/types"
+
```
EasterTheBunny, line 13: I would like to keep any chain specific logic out of the plugin. Capture chain specific logic in the pkg/chain package and only consume common types in the plugin directly.

internal/keepers/service.go
```diff
@@ -53,38 +68,27 @@ func newOnDemandUpkeepService(ratio sampleRatio, registry types.Registry, logger
...
-	size := s.ratio.OfInt(len(keys))
-
-	s.logger.Printf("%d keys selected by provided ratio %s", size, s.ratio)
-	if size <= 0 {
-		return []*types.UpkeepResult{}, nil
-	}
+	s.eligibleUpkeepKeysLock.Lock()
+	keys := make([]types.UpkeepKey, len(s.eligibleUpkeepKeys))
+	copy(keys, s.eligibleUpkeepKeys)
+	s.eligibleUpkeepKeysLock.Unlock()
```
EasterTheBunny, line 85: Instead of using properties of the service, create a data structure that has the single job of maintaining and producing eligible keys. It should have its own internal locks, but should also purge keys like the cache.

internal/keepers/service.go
```diff
@@ -53,38 +68,27 @@ func newOnDemandUpkeepService(ratio sampleRatio, registry types.Registry, logger
...
 
-	if s.workers == nil {
-		panic("cannot sample upkeeps without runner")
+	if len(keys) == 0 {
+		return nil, nil
 	}
 
-	return s.parallelCheck(ctx, keys[:size], filters...)
+	return s.parallelCheck(ctx, keys, filters...)
```
EasterTheBunny, line 91: This should no longer be checking keys. The items from the sampling routine `runSamplingUpkeeps` should populate the cache such that the lines above should provide all the results necessary.

internal/keepers/service.go
```diff
@@ -265,6 +274,71 @@ Outer:
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	ch := make(chan *ethtypes.Header, 1)
+	sub, err := s.headSubscriber.SubscribeNewHead(ctx, ch)
+	if err != nil {
+		return err
+	}
+
+	heads := make(chan *ethtypes.Header)
```
EasterTheBunny, line 287: Extension of previous comment on eth specific content. The header that is returned by the subscriber should be chain generic.

EasterTheBunny (changes requested):

internal/keepers/service.go
begmaroman: But filters are part of the SampleUpkeep function call so we can only pre-populate eligible upkeep keys, actual checks should be done based on filters. Or do you mean checking upkeeps for all keys, and then filter final results?

internal/keepers/service.go
begmaroman: Done

internal/keepers/service.go
EasterTheBunny: You should be able to apply the filters to the list produced by `GetActiveUpkeeps`. From there, just run the key checks.

In the results, you should keep the results from the previous run until the latest run is complete. Both collections should be returned in a `SampleUpkeeps` with filters applied again. This way there is overlap between current and previous head and partial results don't inhibit complete sampling.

begmaroman: Done

internal/keepers/service.go
begmaroman: Done

internal/keepers/factory.go
```diff
@@ -12,32 +12,47 @@ import (
 
 const maxObservationLength = 1_000
 
+var _ types.ReportingPluginFactory = (*keepersReportingFactory)(nil)
```
infiloop2, line 15: curious, what does this statement do? The blank variable can't be used anywhere?

pkg/chain/evmclient.go
```diff
@@ -69,6 +61,27 @@ func (ec *evmClient) BatchCallContext(ctx context.Context, b []rpc.BatchElem) er
...
+			return ctx.Err()
+		case <-sub.Err():
+			if sub, err = ec.SubscribeNewHead(ctx, ch); err != nil {
+				return err
+			}
+		case head := <-ch:
+			cb(types.BlockKey(head.Number.String()))
+		}
+	}
+}
```
infiloop2, line 83: q for my understanding: in prod this code is not actually used, we use the evm client from main CL repo right?

internal/keepers/service.go
```diff
@@ -132,15 +143,82 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
-	samples := newSyncedArray[*types.UpkeepResult]()
+func (s *onDemandUpkeepService) runSamplingUpkeeps() error {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	heads := make(chan types.BlockKey)
+	defer close(heads)
+
+	// Start the sampling upkeep process for heads
+	go func() {
+		for head := range heads {
```
infiloop2, line 166: Can we just skip to the latest head here?

infiloop2: Also i might be missing things, do we listen to the heads channel somewhere?

internal/keepers/factory.go
begmaroman: This is the interface compatibility check

pkg/chain/evmclient.go
begmaroman: This is the wrapper that allows subscribing on new heads in more chain-agnostic way

internal/keepers/service.go
begmaroman: Nope, we are going over all passed heads. It is always the latest one.

internal/keepers/service.go
```diff
@@ -132,15 +143,82 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
+			if len(keys) == 0 {
+				s.samplingResults.purge()
+				continue
+			}
+
+			// select x upkeeps at random from set
+			keys = s.shuffler.Shuffle(keys)
+			size := s.ratio.OfInt(len(keys))
+
+			s.logger.Printf("%d keys selected by provided ratio %s", size, s.ratio)
```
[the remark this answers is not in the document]
begmaroman: @EasterTheBunny should we check all keys here and don't apply:
```go
keys = s.shuffler.Shuffle(keys)
size := s.ratio.OfInt(len(keys))
```
?

internal/keepers/factory.go
EasterTheBunny: It ensures the keepersReportingFactory implements the `types.ReportingPluginFactory`. If not, you can be assured the type system will catch error.

internal/keepers/service.go
begmaroman: Yes, if you look at the logic below, you could find a select statement there with the explanation. Basically, it just skips the incoming head if the current process is busy yet.

pkg/chain/evmclient.go
infiloop2: What i wanted to confirm was that `NewClient` does not seem to be called anywhere in this repo other than mocks / tests. Is that correct?

internal/keepers/service.go
```diff
@@ -132,15 +143,90 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
+	// Cancel context when receiving the stop signal
+	go func() {
+		<-s.stopProcs
+		cancel()
+	}()
+
+	return s.headSubscriber.OnNewHead(ctx, func(blockKey types.BlockKey) {
+		// This is needed in order to do not block the process when a new head comes in.
+		// The running upkeep sampling process should be finished first before starting
+		// sampling for the next head.
```
infiloop2, line 180: question about channel: when we are doing `s.handleIncomingHead(ctx, head)` for a particular head, is it removed from the channel or will it be removed once that call finishes? 

What i want to confirm is that if this behaviour is possible

- Head no. 10 is put into channel
- handleIncomingHead(10) is called by the other go routine
- Head no 11. is put into the channel (Not sure about this)
- Head no 12. comes and is skipped since channel is full
- Head no 13. comes and is skipped since channel is full
- handleIncomingHead(10) completes
- handleIncomingHead(11) is called

internal/keepers/service.go
```diff
@@ -132,15 +143,90 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
+		select {
+		case heads <- blockKey:
+		default:
+		}
+	})
+}
+
+// handleIncomingHead performs checking upkeep logic for all eligible keys of the given head
+func (s *onDemandUpkeepService) handleIncomingHead(ctx context.Context, head types.BlockKey) {
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
infiloop2, line 190: Have we considered an alternative where we cancel the context once we get a new head?

It has some implications on fast chains that we'll end up cancelling quickly but can try to mitigate that by having a 'minimum' time for processing head. The benefit I would see is that we will process upkeeps on fresh head as soon as we receive it.

cc @EasterTheBunny

internal/keepers/service.go
begmaroman: Exactly how it works right now since the head channel has buffer == 1. If I make this channel sync (buffer == 0), then this scenario would be changed to:

- Head no. 10 is put into channel
- handleIncomingHead(10) is called by the other go routine
- Head no 11. comes and is skipped since channel is full (reader is not able to read yet)
- Head no 12. comes and is skipped since channel is full (reader is not able to read yet)
- handleIncomingHead(10) completes
- Head no 13. comes and handleIncomingHead(13) is called

internal/keepers/service.go
begmaroman: This is what we have right now. We don't cancel the running job on a new head.

pkg/chain/evmclient.go
begmaroman: NewEVMClient could be used by plugin's users, i.e. chainlink node.

internal/keepers/service.go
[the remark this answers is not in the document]
EasterTheBunny: We should still apply the sample ratio when checking keys to provide a maximum amount of work a node should do. This makes the work load predictable for node ops.

pkg/chain/evmclient.go
EasterTheBunny: @infiloop2 some of these constructors are used in the `chainlink` repo to build and pass dependencies to the plugin.

EasterTheBunny: It helps separate chain specific logic from the plugin.

internal/keepers/service.go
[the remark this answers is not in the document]
infiloop2: hmm, not really sure if it makes it predictable: as num upkeeps increase the load will increase. 
Also the current parameter of target in round no longer makes sense.


I think in practice we are always going to use 0.999 probability. Might be fine to leave this as is for now and we can do a cleanup of configs later

internal/keepers/service.go
infiloop2: ok i think we need to design this better. 

In first case we'll end up checking on old block (block 11) once `handleIncomingHead` finishes
In second case we will have to wait for a new block once `handleIncomingHead` finishes

Both aren't ideal

internal/keepers/service.go
infiloop2: What would be the tradeoffs for both approaches? We should think through this

infiloop2 (changes requested):
Requesting changes to think through more on the design on how we handle heads when previous one is being processed

internal/keepers/service.go
infiloop2: also the 10 seconds seems arbitrary. Should think through it

EasterTheBunny: I don't think we should cancel when a new head comes in. It doesn't provide any value to start over for every head. We can just run again with the latest head even with some gaps. This is a lot like what we're going to be doing in observability due to the same RPC latency issues we see there.

internal/keepers/service.go
EasterTheBunny: @begmaroman maybe a new struct would be useful that behaves a lot like Go's implementation of `Ticker`. That is, if a new head has come in and the receiver is slow to consume, older heads are dropped and the latest is always waiting. If the consumer is fast, the cadence is just that of incoming heads.

internal/keepers/service.go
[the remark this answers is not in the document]
EasterTheBunny: There just isn't a need for all nodes to check all keys. If we have 1000 keys with 8 nodes, each node only needs to check ~400 keys. Not applying the ratio selection just adds unnecessary load.

I want to make a pass later on config cleanup when we're done making major changes.

internal/keepers/factory.go
```diff
@@ -15,10 +15,11 @@ const maxObservationLength = 1_000
 var _ types.ReportingPluginFactory = (*keepersReportingFactory)(nil)
 
 type ReportingFactoryConfig struct {
-	CacheExpiration       time.Duration
-	CacheEvictionInterval time.Duration
-	MaxServiceWorkers     int
-	ServiceQueueLength    int
+	SamplingUpkeepJobDuration time.Duration
```
EasterTheBunny, line 18: This should be in the off chain config instead of the reporting factory config. The upkeep job duration is a configuration we want to control across the entire network.

EasterTheBunny (changes requested):

internal/keepers/service.go
```diff
@@ -132,25 +146,101 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
+		// sampling for the next head.
+		select {
+		case heads <- blockKey:
+		default:
+		}
+	})
+}
+
+// handleIncomingHead performs checking upkeep logic for all eligible keys of the given head
+func (s *onDemandUpkeepService) handleIncomingHead(ctx context.Context) {
```
infiloop2, line 192: nit. can we rename this function to indicate it doesn't really use the incoming head, instead always uses the latest? 

maybe `processLatestHead` of something?

EasterTheBunny (approved):

PR #42, opened 2022-12-06
Enable batching of upkeeps in a single report

begmaroman:
Updated reporting logic in order to perform multiple upkeeps within one perform


internal/keepers/ocr.go
```diff
@@ -100,38 +100,67 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+	toPerform := make([]ktypes.UpkeepResult, 0, len(upkeepsInfo))
+	for i, upkeepInfo := range upkeepsInfo {
+		upkeepResult, ok := keyToUpkeep[string(eligibleUpkeeps[i])]
+		if !ok {
+			return false, nil, fmt.Errorf("upkeep with key %s not found in the info list", string(eligibleUpkeeps[i]))
+		}
+
+		if reportCapacity+upkeepInfo.ExecuteGas > k.reportGasLimit {
+			break
+		}
```
infiloop2, line 158: we'll need to account for gas overhead also somewhere. Currently if upkeep gas limit is 5M, then we send txs with 5.3M. 
300K overhead per upkeep is a good margin to add (can probably be tightened as incremental upkeeps don't need full 300K overhead, but should be good to start with)

internal/keepers/ocr.go
```diff
@@ -100,38 +100,67 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+	if len(eligibleUpkeeps) == 0 {
+		k.logger.Printf("OCR report completed successfully with no eligible upkeeps: %s", lCtx)
 		return false, nil, nil
 	}
 
+	// Get upkeep gas limit
+	upkeepsInfo, err := k.service.GetUpkeep(ctx, eligibleUpkeeps...)
+	if err != nil {
+		return false, nil, fmt.Errorf("%w: failed to get upkeeps info: %s", err, lCtx)
+	}
```
infiloop2, line 145: Will it be too complex if we try to batch this call within checkUpkeep itself? Would save us sequential RPC calls.

internal/keepers/factory.go
```diff
@@ -127,5 +128,6 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 			offChainCfg.MinConfirmations,
 			d.logger,
 		),
+		reportGasLimit: offChainCfg.GasLimitPerReport,
```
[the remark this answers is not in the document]
begmaroman: Where do we have it on the plugin side? 
I saw one on the chainlink node side. As I understand, we will configure it on the chainlink node side as an off-chain config.

internal/keepers/ocr.go
begmaroman: The gas overhead should be configured on the chainlink node side where transmitter is implemented

internal/keepers/service.go
```diff
@@ -131,7 +143,56 @@ func (s *onDemandUpkeepService) CheckUpkeep(ctx context.Context, keys ...types.U
...
+
+	// get upkeep info at block number in key
+	getResults, err := s.registry.GetUpkeep(ctx, nonCachedKeys...)
+	if err != nil {
+		return nil, fmt.Errorf("%w: service failed to get upkeep from registry", err)
+	}
+
+	// Cache results
+	for i, u := range getResults {
+		s.getCache.Set(string(keys[nonCachedKeysIdxs[i]]), u, util.DefaultCacheExpiration)
```
infiloop2, line 195: Not sure if we should use cache here. This will result in more time a new gas limit takes to be reflected in our system

infiloop2: Unless we listen for gas limit updated logs and clear the cache

begmaroman: This is the cache per upkeep per block so there shouldn't be any issues with this.

internal/keepers/factory.go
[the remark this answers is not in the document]
infiloop2: there's one which is used here https://github.com/smartcontractkit/chainlink/blob/develop/core/services/relay/evm/ocr2keeper.go#L74 within transmitter (not the plugin)

> we will configure it on the chainlink node side as an off-chain config

offchainConfig is set on the contract so it's a different place than chainlink node config

internal/keepers/ocr.go
infiloop2: yes, but currently since transmitter does not take dynamic gas limits, it always uses a fixed one here we will configure it on the chainlink node side as an off-chain config

internal/keepers/ocr.go
infiloop2: yeah. Maybe it's ok for now and in next version we return the gas limit within simulatePerformUpkeep. wdyt @EasterTheBunny ?

internal/keepers/service.go
infiloop2: yup, cache per block should be fine

internal/keepers/factory.go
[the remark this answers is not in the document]
begmaroman: Right, but as I understand, we don't have access to this value from the plugin itself. We would need additionally pass it on the chainlink node side.

internal/keepers/ocr.go
begmaroman: Ok, so what do you suggest? We have nothing to do with the gas overhead here IMO. This logic is needed to calculate the report capacity.

internal/keepers/ocr.go
```diff
@@ -100,38 +100,67 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+		k.logger.Printf("OCR report completed successfully with no eligible keys: %s", lCtx)
+		return false, nil, nil
+	}
 
-		// No upkeeps found for the given key
-		if len(upkeeps) == 0 {
-			continue
-		}
+	// Check all upkeeps from the given observation
+	checkedUpkeeps, err := k.service.CheckUpkeep(ctx, keys...)
```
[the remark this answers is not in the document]
begmaroman: Hmmm, this is what we already have now. Do you think we should cover this scope within this PR as well?

internal/keepers/factory.go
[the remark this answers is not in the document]
EasterTheBunny: @infiloop2 you're talking about the `OCR2AutomationGasLimit` in the node config, right? What does this currently do and why is it configurable per node? I think we need gas limits in the offChainConfig because all nodes need to generate reports deterministically. Each node having different upper thresholds would cause reports to not match.

internal/keepers/ocr.go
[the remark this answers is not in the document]
EasterTheBunny: @begmaroman I noticed that batch size is defined both in the sampling and in the registry implementation. I think we should make the assumption that batch size should be determined outside of the registry for now. In this case, @infiloop2 , it would make sense to send all keys as a single batch since we're unlikely to be sending more than 100 during the report phase anyway. I do like the idea of shaving off a little time from the context to ensure we have enough time for post processing. If we do that, post processing should be broken out to a different function and paired with a benchmark test. That way we can measure the typical processing time and use that value in the context timeout decision.

internal/keepers/ocr.go
EasterTheBunny: I think we should think through this one a bit more. My big concern is that checking upkeeps and getting upkeeps (if they are both returning different data) are occurring in sequence instead of parallel. Let's go back to what data we need exactly, why it's different from checking upkeeps, and see if we can put that data call somewhere else.

internal/keepers/factory.go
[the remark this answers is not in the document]
infiloop2: @EasterTheBunny yes. Currently this is passed to the transmitter and used as the gas limit for all transmit txs. Agree it is set per node and not optimal as different nodes can technically set that to be different. 

Ideally `OCR2AutomationGasLimit` would also be set through offchainConfig but right now there's no way for the plugin to tell the per report gas limit to the transmitter. That probably requires a libocr change (needs to be dug into)

EasterTheBunny (changes requested):
Decisions:

1. Batch size should be controlled outside of the registry implementation for now. Assume that the keys passed to the registry will be a single batch.
2. The gas limit needs to be coordinated across all nodes. Not sure how nodes use the current gas limit config param. Needs some research.
3. Upkeep gas limit is retrieved from extra eth call. This needs to be done in parallel with check upkeep and simulate upkeep when building a report, not sequentially. Maybe we need to restructure the registry implementation? This probably needs a conversation to find the best method.

internal/keepers/factory.go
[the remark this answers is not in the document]
EasterTheBunny: Hmmm. We might need to live with both for now until we can find a way to consolidate them. That just makes setting the off chain config values more complex and confusing.

internal/keepers/ocr.go
infiloop2: So we need to account for the gas overheads somewhere, can discuss the best place to put that.

From contract side, we need to set the gas limit to be sum of upkeeps gas limits + overhead. Overhead required scales with number of upkeeps, 300K per upkeep is an upper bound

internal/keepers/ocr.go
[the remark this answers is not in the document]
infiloop2: just had an offline conversation with @begmaroman . This is complicated further as we are doing a rpc call for get gas limits after this step, so a simple context timeout shaving would not work at least in the current implementation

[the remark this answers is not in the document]
EasterTheBunny: I think there is already a benchmark test for the OCR report function and the registry and service are both mocked. You might be able to just use the values from that benchmark instead.

internal/keepers/ocr.go
RyanRHall: do we have any plans to make this service stateful? gasLimit seems like a prime contender for a piece of data that could live in a db. also agreed that returning from simulatePerform sounds useful.

internal/keepers/ocr.go
EasterTheBunny: Ahhhh. I get it. The overhead depends on the number of upkeeps. Not just the general overhead. That could be broken down to two different 'overheads': the per upkeep overhead and the general transmit overhead. Though it may not be worth our efforts to split them like this.

internal/keepers/ocr.go
EasterTheBunny: @RyanRHall yes and that's something I'm working on a bit in the chainlink repo. Roman and I are going to discuss it today.

EasterTheBunny (approved):
Looks good!

PR #45, opened 2022-12-08
Add logging when PerformUpkeep log is received

RyanRHall:


pkg/types/types.go
```diff
@@ -57,9 +58,11 @@ type PerformLogProvider interface {
 }
 
 type PerformLog struct {
-	Key           UpkeepKey
-	TransmitBlock BlockKey
-	Confirmations int64
+	Key             UpkeepKey
+	TransmitBlock   BlockKey
+	Confirmations   int64
+	TransactionHash common.Hash
```
EasterTheBunny, line 64: One goal of the plugin is to not introduce any chain specific dependencies. Either make the transaction hash a native type or create a wrapping type for it.

EasterTheBunny (changes requested):

pkg/types/types.go
```diff
@@ -57,9 +57,11 @@ type PerformLogProvider interface {
...
 
 type PerformLog struct {
-	Key           UpkeepKey
-	TransmitBlock BlockKey
-	Confirmations int64
+	Key             UpkeepKey
+	TransmitBlock   BlockKey
+	Confirmations   int64
+	TransactionHash string
+	BlockNumber     int64
```
infiloop2, line 64: We also need to ensure these get populated.  @RyanRHall @EasterTheBunny 

Also is blockNumber the same as transmit block? do we need both?

RyanRHall: blockNumber should indicate the blocknumber that includes the tx which produced the logs - different from transmit block

infiloop2: I think that is the transmit block number. As far as i understand there is checkBlockNumber: block at which check was done, transmitBlockNumber: block at which transmit was confirmed. cc @EasterTheBunny to correct me

PR #49, opened 2022-12-16
POC-ccip-read

cmalec:


pkg/chain/evmregistry_test.go
```diff
@@ -17,6 +20,60 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/pkg/types"
 )
 
+func TestOffchainLookup(t *testing.T) {
```
cmalec, line 23: this test does the sme flow of calls as the node

pkg/chain/evmregistry.go
```diff
@@ -131,6 +144,73 @@ func (r *evmRegistryv2_0) check(ctx context.Context, key types.UpkeepKey, ch cha
...
+			ch <- outStruct{
+				ur:  types.UpkeepResult{},
+				err: err,
+			}
+			return
+		} else {
+			logger.Printf("upkeepInfo: %+v", upkeepInfo)
+			logger.Printf("opts: %+v", opts)
+
+			offchainLookup, err := r.callTargetCheckUpkeep(upkeepInfo, opts, logger)
```
cmalec, line 164: however when i do the same call here it returns 'nil/default' data.

pkg/chain/evmregistry_test.go
cmalec: this test works and does everything as expected

cmalec:
closing POC outdated. we have changed how the api request mechanism will work.

PR #51, opened 2023-01-10
Added logs to the batch creating process

begmaroman:
- Added logs to the batch report building process
- Added offchain config log
- Moved eth_call batch size to const

### DIFF
--- a/internal/keepers/factory.go
+++ b/internal/keepers/factory.go
@@ -60,6 +60,8 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 		return nil, types.ReportingPluginInfo{}, fmt.Errorf("%w: failed to decode off chain config", err)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/keepers/factory.go
```diff
@@ -60,6 +60,8 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 		return nil, types.ReportingPluginInfo{}, fmt.Errorf("%w: failed to decode off chain config", err)
 	}
 
+	d.logger.Printf("successfully decoded offchain config when creating plugin: %s", string(c.OffchainConfig))
```
infiloop2, line 63: will a string conversion like this work directly? is there a better way to print a struct

infiloop2: you can use %v and the struct directly

begmaroman: The initial approach would just print a JSON string. I'm fine with both. Seems like you already pushed the update so can we merge it?

infiloop2 (approved):

PR #53, opened 2023-01-12
solution to dedupe upkeep ids

EasterTheBunny:


internal/keepers/util.go
```diff
@@ -60,7 +60,11 @@ func (_ *cryptoShuffler[T]) Shuffle(a []T) []T {
 type sortUpkeepKeys []ktypes.UpkeepKey
 
 func (s sortUpkeepKeys) Less(i, j int) bool {
-	return string(s[i]) < string(s[j])
+	// TODO: hacky string split assuming separator
```
infiloop2, line 63: can you create a ticket here to fix this.

Also this is same as https://github.com/smartcontractkit/ocr2keepers/blob/main/internal/keepers/filter.go#L80-L81 so would be better to make a common function which does this. (getBlockNumber(upkeepKey), getUpkeepID(upkeepKey))

internal/keepers/util.go
```diff
@@ -138,6 +142,29 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
 		return nil, fmt.Errorf("%w: observation dedupe", err)
 	}
 
+	// TODO: another hacky solution assuming upkeep key structure
```
infiloop2, line 145: same comment. let's collect all places where we do the split of key into single function
and document the hack in a ticket and link it here

infiloop2 (changes requested):
Add comments about better collecting and documenting the hacks

Also could you add a test for this

internal/keepers/util.go
EasterTheBunny: https://smartcontract-it.atlassian.net/browse/AUTO-1480

infiloop2: nit. ideally link them in the code so that anyone who is browsing the repo can get context

EasterTheBunny: dun

infiloop2: todo: remove

internal/keepers/util.go
```diff
@@ -138,6 +143,30 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
...
+	// removes duplicate upkeep ids in preferance of ids at higher blocks
+	// needs to be refactored
+	// AUTO-1480
+	sort.Sort(sortUpkeepKeys(keys))
```
infiloop2, line 150: rm line 150

infiloop2 (approved):

PR #55, opened 2023-01-13
Added report batch size to the offchain config

begmaroman:


pkg/types/types.go
```diff
@@ -128,6 +128,9 @@ type OffchainConfig struct {
 
 	// GasOverheadPerUpkeep is gas overhead per upkeep taken place in the report.
 	GasOverheadPerUpkeep uint32 `json:"gasOverheadPerUpkeep"`
+
+	// ReportBatchSize is the max batch size of the OCR2 report.
+	ReportBatchSize int `json:"reportBatchSize"`
```
infiloop2, line 133: can we rename this maxUpkeepBatchSize

infiloop2 (changes requested):
The idea for this config was to limit the number of upkeeps batched here https://github.com/smartcontractkit/ocr2keepers/blob/main/internal/keepers/ocr.go#L126-L145

pkg/types/types.go
begmaroman: Done

internal/keepers/service.go
```diff
@@ -13,9 +13,6 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/pkg/types"
 )
 
-// keyBatchSize is the value of max items in the eth_call batch
-const keyBatchSize = 10
```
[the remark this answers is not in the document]
begmaroman: My bad, mixed up places. Fixed.

begmaroman:
@infiloop2 mixed up these to batching places. Fixed.

internal/keepers/service.go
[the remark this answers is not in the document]
EasterTheBunny: The plugin is responsible for batch sizing instead of the registry. It would probably be good to make it configurable though.

internal/keepers/service.go
```diff
@@ -284,7 +284,7 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
 
 	// Create batches from the given keys.
 	// Max keyBatchSize items in the batch.
-	keysBatches := createBatches(keysToSend, keyBatchSize)
+	keysBatches := createBatches(keysToSend, maxWorkersBatchSize)
```
EasterTheBunny, line 287: Looks like it still isn't coming from the config.

internal/keepers/service.go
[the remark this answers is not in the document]
infiloop2: that can be a separate ticket

internal/keepers/service.go
begmaroman: It should not come from the config. This is the batch of the worker, but we are talking about this one: https://github.com/smartcontractkit/ocr2keepers/blob/main/internal/keepers/ocr.go#L126-L145

infiloop2 (approved):

PR #58, opened 2023-01-24
Change the UpkeepKey type to an interface

ferglor:
~~In this PR, I've added a BlockKeyAndUpkeepID function to the UpkeepKey type. This function performs a string split on `|` and returns a block key and an upkeep ID as a pair of values.~~

~~I was going to add two separate functions, one for each value, but given that both values are often used together, I thought returning the values as a pair made sense.~~

~~If we ever change the UpkeepKey type to encompass more than two values, it would probably make sense to have separate functions for each value, depending on how the client code uses them.~~

In this PR, I've updated the UpkeepKey type to be an interface with a BlockKeyAndUpkeepID function. The UpkeepKey interface is also now a Stringer, and also supports JSON unmarshalling.

I've added an implementation of this interface to the `chain` package. The BlockKeyAndUpkeepID function returns a BlockKey and an UpkeepIdentifier, though these are often cast to strings in the client code.

A significant number of tests have been updated to use the new chain.UpkeepKey implementation in place of the types.UpkeepKey interface, as these tests instantiate UpkeepKeys as part of their test cases.

### DIFF
--- a/internal/keepers/filter_test.go
+++ b/internal/keepers/filter_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/smartcontractkit/ocr2keepers/pkg/chain"
+
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 
@@ -29,10 +31,10 @@ func TestReportCoordinator(t *testing.T) {
 	}
 
 	// set up the mocks and mock data
-	key1Block1 := types.UpkeepKey("1|1")
-	key1Block2 := types.UpkeepKey("2|1")
-	key1Block3 := types.UpkeepKey("3|1")
-	key1Block4 := types.UpkeepKey("4|1")
+	key1Block1 := chain.UpkeepKey("1|1")
+	key1Block2 := chain.UpkeepKey("2|1")
+	key1Block3 := chain.UpkeepKey("3|1")
+	key1Block4 := chain.UpkeepKey("4|1")
 	id1 := types.UpkeepIdentifier("1")
 	bk2 := types.BlockKey("2")
 	bk3 := types.BlockKey("3")
@@ -255,5 +257,5 @@ func TestReportCoordinator(t *testing.T) {
 
 func assertFilter(t *testing.T, reg *types.MockRegistry, key types.UpkeepKey, id types.UpkeepIdentifier, exp bool, f func(types.UpkeepKey) bool) {
 	reg.Mock.On("IdentifierFromKey", key).Return(id, nil)
-	assert.Equal(t, exp, f(key), "filter should return '%v' to indicate key should not be filtered out at block %s", exp, string(key))
+	assert.Equal(t, exp, f(key), "filter should return '%v' to indicate key should not be filtered out at block %s", exp, key)
 }
--- a/internal/keepers/filter_test.go
+++ b/internal/keepers/filter_test.go
@@ -6,12 +6,11 @@ import (
 	"testing"
 	"time"
 
-	"github.com/smartcontractkit/ocr2keepers/pkg/chain"
-
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 
 	"github.com/smartcontractkit/ocr2keepers/internal/util"
+	"github.com/smartcontractkit/ocr2keepers/pkg/chain"
 	"github.com/smartcontractkit/ocr2keepers/pkg/types"
 )
 
--- a/internal/keepers/ocr.go
+++ b/internal/keepers/ocr.go
@@ -227,7 +227,7 @@ func (k *keepers) ShouldTransmitAcceptedReport(_ context.Context, rt types.Repor
 		// (while others may not have), don't try to transmit again
 		// TODO: reevaluate this assumption
 		if transmitConfirmed {
-			k.logger.Printf("not transmitting report because upkeep '%s' was already transmitted: %s", string(id.Key), lCtx)
+			k.logger.Printf("not transmitting report because upkeep '%s' was already transmitted: %s", id.Key, lCtx)
 			return false, nil
 		}
 
--- a/internal/keepers/util.go
+++ b/internal/keepers/util.go
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/smartcontractkit/ocr2keepers/pkg/chain"
+
 	"github.com/smartcontractkit/libocr/offchainreporting2/types"
 	"github.com/smartcontractkit/ocr2keepers/internal/util"
 	ktypes "github.com/smartcontractkit/ocr2keepers/pkg/types"
@@ -60,7 +62,7 @@ func (_ *cryptoShuffler[T]) Shuffle(a []T) []T {
 type sortUpkeepKeys []ktypes.UpkeepKey
 
 func (s sortUpkeepKeys) Less(i, j int) bool {
-	return string(s[i]) < string(s[j])
+	return s[i].String() < s[j].String()
 }
 
 func (s sortUpkeepKeys) Swap(i, j int) {
@@ -129,14 +131,19 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
 		}
 
 		var ob []ktypes.UpkeepKey
-		err = decode(b, &ob)
 
 		// a single observation returning an error here can void all other
 		// good observations. ensure this loop continues on error, but collect
 		// them and throw an error if ALL observations fail at this point.
+		var keys []chain.UpkeepKey
+		err = decode(b, &keys)
 		if err != nil {
 			parseErrors++
 			continue
+		}
+
+		for _, o := range keys {
+			ob = append(ob, o)
 		}
 
 		sort.Sort(sortUpkeepKeys(ob))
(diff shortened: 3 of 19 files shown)


pkg/types/types.go
```diff
@@ -72,6 +73,14 @@ type Address []byte
 // upkeep at a block number
 type UpkeepKey []byte
 
+func (u UpkeepKey) BlockKeyAndUpkeepID() (blockKey string, upkeepID string) {
```
ferglor, line 76: I was thinking of calling this function something like `Explode` or `Components` or `IDs`, but I thought the longer, more descriptive name made it easier for the client functions to use

EasterTheBunny, line 76: This function should not return strings. The return types should be `BlockKey` and `UpkeepIdentifier`

EasterTheBunny (changes requested):

pkg/types/types.go
ferglor: Okay cool; a lot of the client code that uses this functionality expects these values as strings, are we okay with casting the BlockKey and UpkeepIdentifier any time they're needed?

pkg/types/types.go
```diff
@@ -72,6 +73,14 @@ type Address []byte
 // upkeep at a block number
 type UpkeepKey []byte
```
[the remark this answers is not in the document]
ferglor: If we move the implementation to the chainlink repo, how should we deal with instantiating an UpkeepKey implementation within the this repo, e.g.: https://github.com/smartcontractkit/ocr2keepers/blob/main/pkg/chain/evmreports.go#L186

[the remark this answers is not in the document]
ferglor: Does it maybe make sense to have the UpkeepKey interface and implementations live in libocr? 🤔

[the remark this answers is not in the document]
EasterTheBunny: To allow the simulators to continue functioning, it might make sense to put the implementation in `pkg/chain` for now. This way the `chainlink` repo can access it, implementations can be chain specific, and the simulators can use the implementations. 

`libocr` should only be general OCR stuff. The UpkeepKey is specific to the plugin.

pkg/types/types.go
EasterTheBunny: I think that is fine. I would prefer using distinct types for BlockKey and UpkeepIdentifier because functions that should expect them as inputs will eventually benefit from the type references as each morphs into interfaces. Casting should be fine until we create helper functions such as `BlockKey.After(BlockKey)`. These will likely be needed later.

pkg/chain/upkeep_key.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"encoding/json"
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+	"strings"
+)
+
+type UpkeepKey []byte
+
+func (u UpkeepKey) BlockKeyAndUpkeepID() (types.BlockKey, types.UpkeepIdentifier) {
```
EasterTheBunny, line 11: Judging by the findings of the recent audit, this should probably return an error as well.

pkg/types/types.go
```diff
@@ -70,7 +70,10 @@ type Address []byte
 
 // UpkeepKey is an identifier of an upkeep at a moment in time, typically an
 // upkeep at a block number
-type UpkeepKey []byte
+type UpkeepKey interface {
+	BlockKeyAndUpkeepID() (BlockKey, UpkeepIdentifier)
```
EasterTheBunny, line 74: json marshaling should probably also be included here.

EasterTheBunny (changes requested):
Only a couple more changes and we can get it merged. :)

pkg/types/types.go
ferglor: Do you mean, add a custom marshaling function to the concrete type, or add the json.Unmarshaler interface to this interface?

internal/keepers/filter_test.go
```diff
@@ -1,6 +1,7 @@
 package keepers
 
 import (
+	"github.com/smartcontractkit/ocr2keepers/pkg/chain"
```
begmaroman, line 4: I suggest to add goimports tool to your file watchers in IDE

pkg/chain/upkeep_key.go
```diff
@@ -0,0 +1,32 @@
+package chain
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+	"strings"
+)
+
+type UpkeepKey []byte
```
begmaroman, line 10: I would have struct here with block number and upkeep ID fields. It would be a more accurate implementation IMO.

pkg/types/types.go
```diff
@@ -70,7 +70,10 @@ type Address []byte
 
 // UpkeepKey is an identifier of an upkeep at a moment in time, typically an
 // upkeep at a block number
-type UpkeepKey []byte
+type UpkeepKey interface {
+	BlockKeyAndUpkeepID() (BlockKey, UpkeepIdentifier, error)
+	fmt.Stringer
```
begmaroman, line 75: This interface should include json marshaler and unmarshaler interfaces

internal/keepers/filter_test.go
ferglor: I do have goimports added to my file watchers, is there a specific configuration you use for it/would like me to use?

pkg/chain/upkeep_key.go
ferglor: I totally agree, but to me that's out of the scope of this ticket? For this change alone I just wanted to swap out all existing references to the old UpkeepKey type with references to the new UpkeepKey interface

pkg/types/types.go
ferglor: I don't think it should? We should never directly call the marshal or unmarshal functions on an UpkeepKey implementation, and for any other implementations that we introduce, it should be implementation-specific whether or not we need to define custom behavior for marshalling and unmarshalling?

internal/keepers/filter_test.go
begmaroman:
```go
import (
    // built-in imports

    // third-party imports

    // internal imports 
)
```

pkg/chain/upkeep_key.go
begmaroman: I would not merge it with the bytes slice type so the ticket scope could be extended or the parent branch could be created until we have a struct here. Would it be difficult to implement it within this PR?

ferglor: I don't think it would be difficult to implement, it's just that, it was already a byte slice prior to this PR, and I wanted to have the least amount of changes necessary to get the interface changes over the line? So, I want to keep things as-is, without introducing too many changes in one PR, does that make sense?

begmaroman: It makes sense to me but I would really want to change it to struct in the close future

internal/keepers/ocr.go
```diff
@@ -227,7 +227,7 @@ func (k *keepers) ShouldTransmitAcceptedReport(_ context.Context, rt types.Repor
 		// (while others may not have), don't try to transmit again
 		// TODO: reevaluate this assumption
 		if transmitConfirmed {
-			k.logger.Printf("not transmitting report because upkeep '%s' was already transmitted: %s", string(id.Key), lCtx)
+			k.logger.Printf("not transmitting report because upkeep '%s' was already transmitted: %s", id.Key.String(), lCtx)
```
begmaroman, line 230: You don't need to do .String(), it would be done automatically if the type implements stringer interface

internal/keepers/filter_test.go
```diff
@@ -255,5 +256,5 @@ func TestReportCoordinator(t *testing.T) {
 
 func assertFilter(t *testing.T, reg *types.MockRegistry, key types.UpkeepKey, id types.UpkeepIdentifier, exp bool, f func(types.UpkeepKey) bool) {
 	reg.Mock.On("IdentifierFromKey", key).Return(id, nil)
-	assert.Equal(t, exp, f(key), "filter should return '%v' to indicate key should not be filtered out at block %s", exp, string(key))
+	assert.Equal(t, exp, f(key), "filter should return '%v' to indicate key should not be filtered out at block %s", exp, key.String())
```
begmaroman, line 259: You don't need to do .String(), it would be done automatically if the type implements stringer interface

pkg/chain/upkeep_key.go
```diff
@@ -0,0 +1,32 @@
...
+		return types.BlockKey(components[0]), types.UpkeepIdentifier(components[1]), nil
+	}
+	return types.BlockKey(""), types.UpkeepIdentifier(""), fmt.Errorf("%w: missing data in upkeep key", ErrUpkeepKeyNotParsable)
+}
+
+func (u UpkeepKey) String() string {
+	return string(u)
+}
+
+func (u *UpkeepKey) UnmarshalJSON(b []byte) error {
```
begmaroman, line 24: What's the reason for having this function? It seems like it just implements a default []byte bevavior of json marshaling

ferglor: It's actually no longer needed, it was a hangover from testing my changes mid way through refactoring, I've removed it 👍

internal/keepers/util.go
```diff
@@ -129,16 +131,21 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
 		}
 
 		var ob []ktypes.UpkeepKey
-		err = decode(b, &ob)
 
 		// a single observation returning an error here can void all other
 		// good observations. ensure this loop continues on error, but collect
 		// them and throw an error if ALL observations fail at this point.
+		var keys []chain.UpkeepKey
```
EasterTheBunny, line 138: No change, but this is going to come back to bite us. Referencing the distinct struct type here instead of an interface makes the code less adaptable to different chain implementations in the future. I think the only choice we have though is to build the structure into the observation encoding/decoding to decide on decoded struct types. This will need to be a TODO for later.

pkg/types/types.go
EasterTheBunny: I think the thought process I was going down was to enforce custom marshaling/unmarshaling on future implementations. Adding the functions to this interface says to future devs "we use JSON encoding and you need to implement custom functions for it".

EasterTheBunny (changes requested):

pkg/types/types.go
ferglor: Right, but we shouldn't need to *enforce* custom implementations for marshalling or unmarshalling? 

Like, if I add a new UpkeepKey implementation, and I don't like the default marshalling/unmarshalling functionality for that type, then I can specify my own Marshal and Unmarshal functions as desired, which is fine. 

But if we add the json marshal and unmarshal interfaces here, then we're saying that every UpkeepKey type *must* implement their own custom functionality for both operations, even in cases where the "default" marshalling and unmarshalling is the desired behavior. 

So, potentially we'd be forcing future developers to explicitly re-write the default marshalling and unmarshalling behavior for certain types? 

That's what I had in an earlier version of this PR, when I added the json.Unmarshaler interface to this interface, and my unmarshal function was just performing the default unmarshal behavior, which was just extra unnecessary code, (which has since been removed).

EasterTheBunny: Right, we would be forcing future developers to explicitly re-write default marshaling. The intention I'm after is actually validation. I would like to enforce the idea that the incoming upkeep keys need to be validated. Maybe a validation function instead? I was thinking about unmarshaling as the validation step, but it can be done another way.

EasterTheBunny (approved):
This will be a good start to further build on. Looks good to me.

EasterTheBunny (approved):

PR #60, opened 2023-01-25
Fix string comparison for block keys in filter

begmaroman:


internal/keepers/filter.go
```diff
@@ -80,18 +81,16 @@ func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
...
 			if len(parts) == 2 {
-				blKey = parts[0]
+				blKey, _ = strconv.Atoi(parts[0])
 			}
 
-			// only apply filter if key block is after block in cache
-			if len(bl.TransmitBlockNumber) > 0 && blKey > string(bl.TransmitBlockNumber) {
-				return true
-			}
+			transmitBlockNumber, _ := strconv.Atoi(string(bl.TransmitBlockNumber))
```
infiloop2, line 90: there's also a case when `bl.TransmitBlockNumber` can be empty string where we would have returned false, i think we've changed the logic in that case

begmaroman: Done

EasterTheBunny (changes requested):

internal/keepers/filter.go
```diff
@@ -80,18 +81,21 @@ func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
 			// TODO: the key is constructed in the registry. splitting out the
 			// block number here is a hack solution that should be fixed asap.
 			// AUTO-1480
-			var blKey string
+			var blKey int
 			parts := strings.Split(string(key), "|")
 			if len(parts) == 2 {
-				blKey = parts[0]
+				blKey, _ = strconv.Atoi(parts[0])
```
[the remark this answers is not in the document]
begmaroman: Done

internal/keepers/filter.go
```diff
@@ -80,18 +81,21 @@ func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
...
-			// only apply filter if key block is after block in cache
-			if len(bl.TransmitBlockNumber) > 0 && blKey > string(bl.TransmitBlockNumber) {
-				return true
+			// Return false if empty
+			if bl.TransmitBlockNumber == "" {
+				return false
 			}
 
-			return false
+			transmitBlockNumber, _ := strconv.Atoi(string(bl.TransmitBlockNumber))
```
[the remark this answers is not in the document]
begmaroman: Done

EasterTheBunny (approved):

internal/keepers/filter.go
```diff
@@ -80,18 +81,27 @@ func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
...
+				return false
 			}
 
-			return false
+			transmitBlockNumber, err := strconv.Atoi(string(bl.TransmitBlockNumber))
+			if err != nil {
+				return false
+			}
+
+			// only apply filter if key block is after block in cache
```
infiloop2, line 103: nit. comment is opposite. 
key if filtered if key block is less than cache transmit

PR #61, opened 2023-01-25
Refactor OnNewHead to return a channel

EasterTheBunny:
The behavior of `OnNewHead` resulted in a reference back to the calling service by passing a function to the subscription. Converting to a channel eliminates the outside reference to service and only creates references from service to outside entities.

### DIFF
--- a/pkg/chain/evmclient.go
+++ b/pkg/chain/evmclient.go
@@ -19,17 +19,23 @@ import (
 // evmClient expends the base EVM client by splitting batch calls into sub-batches
 type evmClient struct {
 	*ethclient.Client
+	chHead    chan types.BlockKey
 	rpcClient *rpc.Client
 	batchSize int
 }
 
 // NewEVMClient is the constructor of evmClient
 func NewEVMClient(client *rpc.Client, batchSize int) types.EVMClient {
-	return &evmClient{
+	c := &evmClient{
 		Client:    ethclient.NewClient(client),
+		chHead:    make(chan types.BlockKey, 1000),
 		rpcClient: client,
 		batchSize: batchSize,
 	}
+
+	c.subscribe(context.Background())
+
+	return c
 }
 
 // HeaderByNumber returns a block header from the current canonical chain. If number is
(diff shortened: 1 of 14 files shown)


pkg/chain/evmregistry.go
```diff
@@ -17,18 +17,7 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/pkg/types"
 )
 
-const (
-	ActiveUpkeepIDBatchSize int64  = 10000
-	separator               string = "|"
-)
-
```
infiloop2, line 24: is this strange strictly needed? it looks safe as just a move, but trying to minimise the changes in bugfix

pkg/chain/evmclient.go
```diff
@@ -61,23 +67,27 @@ func (ec *evmClient) BatchCallContext(ctx context.Context, b []rpc.BatchElem) er
...
-			return ctx.Err()
+			return
 		case <-sub.Err():
 			if sub, err = ec.SubscribeNewHead(ctx, ch); err != nil {
-				return err
+				return
 			}
 		case head := <-ch:
-			cb(types.BlockKey(head.Number.String()))
+			ec.chHead <- types.BlockKey(head.Number.String())
```
infiloop2, line 90: will this be a blocking call if chHead is full?

pkg/chain/evmclient.go
```diff
@@ -19,17 +19,23 @@ import (
...
 	rpcClient *rpc.Client
 	batchSize int
 }
 
 // NewEVMClient is the constructor of evmClient
 func NewEVMClient(client *rpc.Client, batchSize int) types.EVMClient {
-	return &evmClient{
+	c := &evmClient{
 		Client:    ethclient.NewClient(client),
+		chHead:    make(chan types.BlockKey, 1000),
```
infiloop2, line 31: any reason to not have it 1 length as we had before

EasterTheBunny: Same as above

pkg/chain/evmregistry.go
EasterTheBunny: No. But like the client, the evm registry isn't used by the chainlink node. It is only used by the simulator.

infiloop2 (approved):

PR #64, opened 2023-02-02
Chore/refactor worker group as utility

EasterTheBunny:
The intention of this change is to clean up parallelization into distinct functions that are easier to follow and provide external access to those functions for use in the `chainlink` repo. This code will be helpful for parallelizing upkeep config getter functions.


pkg/util/worker.go
```diff
@@ -143,6 +144,58 @@ func (wg *workerGroup[T]) Do(ctx context.Context, w work[T]) error {
...
+
+func makeJobFunc[T, K any](jobCtx context.Context, value T, jobFunc JobFunc[T, K]) WorkItem[K] {
+	return func(svcCtx context.Context) (K, error) {
+		// the jobFunc should exit in the case that either the job context
+		// cancels or the worker service context cancels. To ensure we don't end
+		// up with memory leaks, cancel the merged context to release resources.
+		ctx, cancel := MergeContextsWithCancel(svcCtx, jobCtx)
+		v, err := jobFunc(ctx, value)
+		cancel()
+		return v, err
```
ferglor, line 199: Could this be written as:

```
ctx, cancel := MergeContextsWithCancel(svcCtx, jobCtx)
defer cancel()

return jobFunc(ctx, value)
```

EasterTheBunny: Yes! Good call

internal/keepers/service.go
```diff
@@ -217,76 +218,21 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
...
-
-	s.logger.Printf("waiting for results to be read")
-	wg.Wait()
-	close(done)
+	pkgutil.RunJobs(
+		ctx,
+		s.workers,
+		createBatches(filteredKeys, maxWorkersBatchSize),
+		wrapWorkerFunc(s.logger, s.registry),
+		wrapAggregate(s.logger, s.cache, &wResults, samples),
```
ferglor, line 234: Would it make sense to move both these wrapper functions into the onDemandUpkeepService, so they both have pointer receivers? That way you can cut down on what params you're sending in?

ferglor: So you'd end up with:

s.wrapWorkerFunc()

And

s.wrapAggregate(&wResults, samples)

internal/keepers/service.go
```diff
@@ -306,67 +252,56 @@ func (s *onDemandUpkeepService) parallelCheck(ctx context.Context, keys []types.
...
+		if cached {
+			cacheHits++
+			if result.State == types.Eligible {
+				samples.Append(result)
 			}
-			w.Done()
-		case <-done:
-			s.logger.Printf("done signal received for job result aggregator")
-			break Outer
+			continue
```
ferglor, line 267: 🏄‍♂️ Personally I think I'd prefer an else here to append to the keysToSend rather than a continue to prevent it? But this is totally cool if you prefer the continue!

internal/keepers/service.go
EasterTheBunny: Makes sense. I was going for easier unit testing, but this approach connects it to the service instead which is where it belongs.

internal/keepers/service.go
EasterTheBunny: This one I'm inconsistent on. According to official Go style, continue is preferred. But it's really just preference, right?

ferglor (approved):

PR #66, opened 2023-02-09
Updated dedupe function to use fmt.Stringer

begmaroman:


internal/keepers/util_test.go
```diff
@@ -30,34 +36,35 @@ func TestCryptoShuffler(t *testing.T) {
...
+		{Name: "Triple No Match", Sets: [][]testStringer{{"1", "2", "3"}, {"4", "5", "6"}, {"7", "8", "9"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5", "6", "7", "8", "9"}},
+		{Name: "Double w/ Single Overlap", Sets: [][]testStringer{{"1", "2", "3"}, {"3", "4", "5"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5"}},
+		{Name: "Double w/ Single Overlap and Gap", Sets: [][]testStringer{{"1", "2", "3", "6", "7"}, {"3", "4", "5"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5", "6", "7"}},
+		{Name: "Double w/ Double Overlap and Gap", Sets: [][]testStringer{{"1", "2", "3", "4", "6", "7"}, {"3", "4", "5"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5", "6", "7"}},
+		{Name: "Triple w/ Single Overlap", Sets: [][]testStringer{{"1", "2", "3"}, {"1", "2", "3"}, {"3", "4", "5"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5"}},
+		{Name: "Triple w/ Single Overlap and Gap", Sets: [][]testStringer{{"1", "2", "4"}, {"1", "2", "3", "4"}, {"3", "4", "5"}}, ExpectedResult: []testStringer{"1", "2", "3", "4", "5"}},
```
ferglor, line 53: 🏄‍♂️ Would there be any value in adding test cases for strings containing letters, or strings composed of only numbers, but prefixed with zeroes? Not necessarily to challenge the changes here, but as an assurance to future developers what the expected outcome should be?

ferglor (approved):
Groovy 👍

internal/keepers/util_test.go
begmaroman: deduct accepts any kind of stringer so it just does not care about values. Or do you mean something else?

PR #68, opened 2023-02-10
Assign median block to keys

ferglor:
In this PR, I'm modifying the dedupe functionality, so that the upkeep keys are modified to use the median block key from all upkeep keys

Given the the following set of upkeep keys:

``` 
[{block1|key1}, {block2|key1}, {block3|key2}, {block2|key3}]
```

The block keys are as follows:

```block1, block2, block2, block3```

The median of which is `block2`. 

The upkeep keys would then be updated to use `block2` as the block key, resulting in the following upkeep keys:

``` 
[{block2|key1}, {block2|key1}, {block2|key2}, {block2|key3}]
```

With the dedupe operation applied, this list is reduced down to the following:

```
[{block2|key1}, {block2|key2}, {block2|key3}]
```

### DIFF
--- a/internal/keepers/util.go
+++ b/internal/keepers/util.go
@@ -155,7 +155,7 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
 		return nil, fmt.Errorf("%w: cannot prepare sorted key list; observations not properly encoded", err)
 	}
 
-	kys = medianBlock(kys)
+	kys = assignMedianBlockToKeys(kys)
 
 	keys, err := dedupe(kys, filters...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: internal/keepers/util.go [hunk 1]
```diff
@@ -198,6 +201,52 @@ func shuffledDedupedKeyList(attributed []types.AttributedObservation, key [16]by
...
+	copy(dataCopy, data)
+
+	sort.Slice(dataCopy, func(i, j int) bool {
+		return dataCopy[i].Cmp(dataCopy[j]) < 0
+	})
+
+	var median *big.Int
+	if l := len(dataCopy); l == 0 {
+		median = big.NewInt(0)
+	} else if l%2 == 0 {
```
ferglor, line 241: The correct way to calculate a median value across an even number of elements is to average the two middle elements, e.g. for a list of 1, 2, 3, 4, the median value would be (2 + 3) / 2 = 2.5. 

In this case, we'd have an invalid block number, but we could round the result up or down to give us a whole number/valid block number. 

The thing I'm not sure on is, should we only consider the set of block keys received as input, as candidates for the median value, or is it okay for the median value to not equal any of the input block keys?

E.g. for this list of block key: 10, 50, 100, 200, the median would be (50+100)/2 = 75. Is it okay for us to use 75 as the median block key, even though 75 wasn't passed in as part of the observations? Or should we instead limit ourselves to the set of input block keys only, e.g. in this case we'd use wither 50 or 100 as the median block key?

EasterTheBunny: I think in practice, the numbers should be close together and not spaced out a lot. In the case of an even number of values, I think it would be just fine if we selected the larger of the two. The primary goal is to eliminate extremes, not to get a perfect median. I would make a note in the code though on why we're not calculating a proper median. This could be confusing for later devs.

PR #69, opened 2023-02-13
Observation should return a limited number of upkeeps

begmaroman:


infiloop2 (changes requested):
Can you please create a corresponding CL repo PR (draft) which uses this version of ocr2keepers to verify tests pass and no other changes are needed

internal/keepers/ocr.go
```diff
@@ -55,7 +63,7 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
 
 	// keyList produces a sorted result so the following reduction of keys
 	// should be more uniform for all nodes
-	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
+	keys := keyList(filterUpkeeps(results, ktypes.Eligible, observationKeysLimit))
```
infiloop2, line 66: not sure if it makes sense to expand scope of filter. why not a separate step after filter?

internal/keepers/ocr.go
```diff
@@ -11,6 +11,14 @@ import (
 	"golang.org/x/crypto/sha3"
 )
 
+const (
+	// observationKeysLimit is the max number of keys that Observation could return.
+	observationKeysLimit = 1
+
+	// reportKeysLimit is the maximum number of upkeep keys returned by the report phase
+	reportKeysLimit = 1
```
infiloop2, line 19: this needs to be 10

internal/keepers/ocr.go
```diff
@@ -94,7 +102,13 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
 
 	// pass the filter to the dedupe function
 	// ensure no locked keys come through
-	keys, err := shuffledDedupedKeyList(attributed, key, k.filter.Filter())
+	keys, err := shuffledDedupedKeyList(
+		attributed,
+		key,
+		observationKeysLimit,
+		reportKeysLimit,
```
infiloop2, line 109: again, i think we are really expanding the scope of shuffle and deduping. having separate steps would be clearer

internal/keepers/ocr.go
begmaroman: Makes sense. Done.

internal/keepers/ocr.go
begmaroman: Fixed

internal/keepers/ocr.go
begmaroman: It would be more tricky since we need to apply these limits in the middle of that func, and in the middle of the loop.

infiloop2: @ferglor is also doing changes in this function which will refactor and break it into multiple pieces. I think the observation limit will then fit nicely there. 

let's wait for his changes and then we can rebase this PR on top of that

PR #70, opened 2023-02-13
clean up config defaults

infiloop2:
- Clean up defaults for all values
- Remove uniqueReports config which was an unnecessary risk
- Add tests for decoding config

Corresponding CL PR: https://github.com/smartcontractkit/chainlink/pull/8440/files

### DIFF
--- a/internal/keepers/factory.go
+++ b/internal/keepers/factory.go
@@ -77,7 +77,7 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 			// with performData of arbitrary length
 			MaxReportLength: 10_000, // TODO (config): pick sane limit based on expected performData size. maybe set this to block size limit or 2/3 block size limit?
 		},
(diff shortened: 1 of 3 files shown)


internal/keepers/factory.go
```diff
@@ -77,7 +77,7 @@ func (d *keepersReportingFactory) NewReportingPlugin(c types.ReportingPluginConf
 			// with performData of arbitrary length
 			MaxReportLength: 10_000, // TODO (config): pick sane limit based on expected performData size. maybe set this to block size limit or 2/3 block size limit?
 		},
-		UniqueReports: offChainCfg.UniqueReports,
+		UniqueReports: false,
```
EasterTheBunny, line 80: Let's leave a comment here on why we're hard-coding this configuration. There is some magic behind the scenes that should be surfaced here.

infiloop2: done

EasterTheBunny (approved):

PR #72, opened 2023-02-14
Add an UpkeepObservation type 

ferglor:
In this PR, I'm adding a new UpkeepObservation type that encodes a block key and a list of upkeep identifiers

The UpkeepObservation is emitted by the observation phase and consumed by the report phase

Previously, we emitted and consumed a list of upkeep keys

To better illustrate this, before, we had:

```[1|567, 1|999, 1|2000]```

Now we have:

```{
"blockKey": "1",
"upkeepIdentifiers": ["567", "999", "2000"]
}```

The key areas impacted by this change are:

- Limited length encoding
- OCR tests
- shuffledDedupedKeyList

### DIFF
--- a/internal/keepers/ocr.go
+++ b/internal/keepers/ocr.go
@@ -57,7 +57,9 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
 	// should be more uniform for all nodes
 	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
 
-	obs := &ktypes.UpkeepObservation{}
+	obs := &ktypes.UpkeepObservation{
+		UpkeepIdentifiers: []ktypes.UpkeepIdentifier{},
+	}
 	// TODO get latest block from head ticker
 
 	if len(keys) > 0 {
--- a/internal/keepers/ocr.go
+++ b/internal/keepers/ocr.go
@@ -57,7 +57,20 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
 	// should be more uniform for all nodes
 	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
 
-	b, err := limitedLengthEncode(keys, maxObservationLength)
+	obs := &ktypes.UpkeepObservation{}
+	// TODO get latest block from head ticker
+
+	if len(keys) > 0 {
+		var identifiers []ktypes.UpkeepIdentifier
+		for _, upkeepKey := range keys {
+			blockKey, upkeepID, _ := upkeepKey.BlockKeyAndUpkeepID()
+			identifiers = append(identifiers, upkeepID)
+			obs.BlockKey = blockKey
+		}
+		obs.UpkeepIdentifiers = identifiers
+	}
+
+	b, err := limitedLengthEncode(obs, maxObservationLength)
 	if err != nil {
 		return nil, fmt.Errorf("%w: failed to encode upkeep keys for observation: %s", err, lCtx)
 	}
--- a/internal/keepers/ocr_test.go
+++ b/internal/keepers/ocr_test.go
@@ -233,9 +233,30 @@ func TestReport(t *testing.T) {
 			ReportGasLimit: 10000000,
 			Ctx:            func() (context.Context, func()) { return context.Background(), func() {} },
 			Observations: []types.AttributedObservation{
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
 			},
 			Checks: checks{
 				K: []ktypes.UpkeepKey{
@@ -254,9 +275,30 @@ func TestReport(t *testing.T) {
 			ReportGasLimit: 10000000,
 			Ctx:            func() (context.Context, func()) { return context.WithTimeout(context.Background(), time.Second) },
 			Observations: []types.AttributedObservation{
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
 			},
 			Checks: checks{
 				K: []ktypes.UpkeepKey{
@@ -275,9 +317,30 @@ func TestReport(t *testing.T) {
 			ReportGasLimit: 10000000,
 			Ctx:            func() (context.Context, func()) { return context.WithTimeout(context.Background(), time.Second) },
 			Observations: []types.AttributedObservation{
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
 			},
 			Checks: checks{
 				K: []ktypes.UpkeepKey{chain.UpkeepKey("1|1")},
@@ -293,9 +356,31 @@ func TestReport(t *testing.T) {
 			ReportGasLimit: 10000000,
 			Ctx:            func() (context.Context, func()) { return context.Background(), func() {} },
 			Observations: []types.AttributedObservation{
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|2"), chain.UpkeepKey("1|1")}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey("1|1")}))},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("2"),
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+						},
+					})),
+				},
 			},
 			Checks: checks{
 				K: []ktypes.UpkeepKey{
@@ -325,9 +410,31 @@ func TestReport(t *testing.T) {
 			ReportGasLimit: 10000000,
 			Ctx:            func() (context.Context, func()) { return context.Background(), func() {} },
 			Observations: []types.AttributedObservation{
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey([]byte("1|2"))}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey([]byte("1|1")), chain.UpkeepKey([]byte("1|2"))}))},
-				{Observation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{chain.UpkeepKey([]byte("1|2"))}))},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("2"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("1"),
+							ktypes.UpkeepIdentifier("2"),
+						},
+					})),
+				},
+				{Observation: types.Observation(mustEncodeUpkeepObservation(
+					&ktypes.UpkeepObservation{
+						BlockKey: ktypes.BlockKey("1"),
+						UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+							ktypes.UpkeepIdentifier("2"),
+						},
+					})),
+				},
 			},
 			FilterOut: []ktypes.UpkeepKey{
 				chain.UpkeepKey("1|1"),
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: pkg/types/types.go [hunk 1]
```diff
@@ -68,6 +68,11 @@ type BlockKey string
 
 type Address []byte
 
+type UpkeepObservation struct {
+	BlockKey          BlockKey           `json:"blockKey"`
+	UpkeepIdentifiers []UpkeepIdentifier `json:"upkeepIdentifiers"`
```
ferglor, line 73: Should we give smaller/more succinct JSON keys for these entries? 🤷‍♂️ It might be the difference of an extra upkeep being included?

internal/keepers/ocr.go
```diff
@@ -57,7 +57,22 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
 	// should be more uniform for all nodes
 	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
 
-	b, err := limitedLengthEncode(keys, maxObservationLength)
+	obs := &ktypes.UpkeepObservation{
+		UpkeepIdentifiers: []ktypes.UpkeepIdentifier{},
+	}
+	// TODO get latest block from head ticker
```
ferglor, line 63: We need to be able to emit an `UpkeepObservation` with a block key and an empty list of upkeep identifiers. Right now, the code relies on a non-empty list of upkeep identifiers to get the block key, but in the scenario where we have an empty list of upkeep identifiers, we need to get the block key from *somewhere*. My understanding is that the HeadTicker should be able to provide this?

infiloop2: evm registry has a function `latestBlock`, we can just expose that https://github.com/smartcontractkit/chainlink/blob/develop/core/services/ocr2/plugins/ocr2keeper/evm/registry.go#L134-L135 ?

internal/keepers/ocr_test.go
```diff
@@ -101,14 +111,27 @@ func TestObservation(t *testing.T) {
...
+				{Key: chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000015")), State: ktypes.Eligible},
 			},
-			ExpectedObservation: types.Observation(mustEncodeKeys([]ktypes.UpkeepKey{
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000001")),
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000002")),
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000003")),
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000004")),
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000005")),
-				chain.UpkeepKey([]byte("100000000000100000000000100000000000100000000000100000000001|100000000000100000000000100000000000100000000006")),
```
ferglor, line 111: I'm not sure I understand why the older version of this test saw the limit reached after the 6th upkeep? Each block key and upkeep ID pair, together with the pipe character, comes in at 105 characters, so I would expect 9 of these to be within the 1KB limit? Not 6? 🤷‍♂️

infiloop2: Encoding is a lot more complex than the linear addition of fields. We ran into that issue while implementing this, and that's why at the end had that code which just encoded the whole thing to confirm the size.

I would recommend just doing an overall encoding and iterating, instead of doing math with number of fields etc

pkg/types/types.go
begmaroman: I would use "block" and "ids"

infiloop2: yup. reduces p2p bandwidth too. I would be ok with 'b' and 'u' also.

EasterTheBunny: I think we're splitting hairs hairs here. We are limiting the number of upkeeps anyway so having one more in an observation is a moot point with the current iteration. Then having descriptive keys is not really useful anyway, because the json encoding is never meant to be human readable anyway. Picking `1` and `2` would be sufficient. So in my opinion, `b` and `u` works just the same as numbering them.

internal/keepers/ocr.go
EasterTheBunny: It would be easiest to add a function to the `upkeepService` interface that exposes the latest block key. The latest block key comes to the service in the the following:

```
func (s *onDemandUpkeepService) start() {
	// TODO: if this process panics, restart it
	go s.cacheCleaner.Run(s.cache)
	go func() {
		ch := s.headSubscriber.HeadTicker()
		for {
			select {
			case <-ch:
				// run with new head
				s.processLatestHead(s.ctx)
			case <-s.ctx.Done():
				return
			}
		}
	}()
}
```

Just keep track of the last known block key from the head ticker channel and return that to the OCR process.

internal/keepers/ocr_test.go
EasterTheBunny: Right. This was tricky. The function does an estimate on the number of keys that can fit in the limit, encodes it, checks the limit, and then walks backward key by key if over the limit. I would be hesitant on iterating one key at a time starting from 1 and encoding at each step before the limit check. I would prefer an optimization technique that reduces the number of encoding rounds.

begmaroman (approved):
Just fix race tests

internal/keepers/ocr.go
infiloop2: worth noting here that the channel was forced to have size 1, the behaviour we wanted was to skip blocks while `s.processLatestHead` was running. Not sure if it causes issues, just something to keep in mind

infiloop2: thinking more, this might be the behaviour we actually want. We do not send the 'latest' head, but the latest head we are checking the upkeeps on. 
This will keep it consistent with the case when we do actually have upkeeps which are eligible

internal/keepers/ocr.go
```diff
@@ -11,6 +11,14 @@ import (
 	"golang.org/x/crypto/sha3"
 )
 
+const (
+	// observationKeysLimit is the max number of keys that Observation could return.
```
infiloop2, line 15: nit. we do not send keys anymore in observation

observationUpkeepLimit is the max number of upkeeps that can be sent in an observation

internal/keepers/ocr.go
```diff
@@ -11,6 +11,14 @@ import (
...
+	observationKeysLimit = 1
+
+	// reportKeysLimit is the maximum number of upkeep keys returned by the report phase
```
infiloop2, line 18: nit. reportKeysLimit is the maximum number of upkeep keys checked during the report phase

internal/keepers/ocr.go
```diff
@@ -57,7 +65,31 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
...
+	obs := &ktypes.UpkeepObservation{
+		BlockKey:          ktypes.BlockKey(latestBlock.String()),
+		UpkeepIdentifiers: []ktypes.UpkeepIdentifier{},
+	}
+
+	if len(keys) > 0 {
+		var identifiers []ktypes.UpkeepIdentifier
+		for _, upkeepKey := range keys {
+			_, upkeepID, _ := upkeepKey.BlockKeyAndUpkeepID()
+			identifiers = append(identifiers, upkeepID)
```
infiloop2, line 87: not getting the blockKey from the key? 
We'll need to ensure the latestBlock we send matches the upkeep keys

internal/keepers/ocr.go
```diff
@@ -94,7 +126,7 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
 
 	// pass the filter to the dedupe function
 	// ensure no locked keys come through
-	keys, err := shuffledDedupedKeyList(attributed, key, k.filter.Filter())
+	keys, err := shuffleUniqueObservations(attributed, key, reportKeysLimit, k.filter.Filter())
```
infiloop2, line 129: can we refactor this please.

1. Separate function to get median and generate new keys
2. split out reportKeysLimit as a separate step

internal/keepers/ocr.go
begmaroman: Fixed

internal/keepers/ocr.go
begmaroman: Done

internal/keepers/ocr.go
begmaroman: Done

internal/keepers/util.go
```diff
@@ -110,67 +111,51 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
+	upkeepKeys, err := observationsToUpkeepKeys(observations)
+	if err != nil {
+		return nil, err
+	}
 
-		sort.Sort(sortUpkeepKeys(ob))
-		kys[i] = ob
+	if len(upkeepKeys) > totalLimit {
+		upkeepKeys = upkeepKeys[:totalLimit]
 	}
```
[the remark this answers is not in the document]
begmaroman: Done in a little bit different way

pkg/types/types.go
```diff
@@ -131,6 +137,9 @@ type OffchainConfig struct {
 
 	// MaxUpkeepBatchSize is the max upkeep batch size of the OCR2 report.
 	MaxUpkeepBatchSize int `json:"maxUpkeepBatchSize"`
+
+	// ReportBlockLag is the number to subtract from median block number during report phase.
+	ReportBlockLag int `json:"reportBlockLag"`
```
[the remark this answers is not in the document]
begmaroman: Done

internal/keepers/ocr.go
infiloop2: discussed offline. will get the latest block from sampling process

internal/keepers/service.go
```diff
@@ -178,37 +181,31 @@ func (s *onDemandUpkeepService) processLatestHead(ctx context.Context) {
...
-	}
 
 	// select x upkeeps at random from set
 	keys = s.shuffler.Shuffle(keys)
-	size := s.ratio.OfInt(len(keys))
+	sampleSize := s.ratio.OfInt(len(keys))
 
-	s.logger.Printf("%d results selected by provided ratio %s", size, s.ratio)
-	if size <= 0 {
-		s.samplingResults.purge()
```
ferglor, line 200: As part of the offline discussion with @infiloop2, we decided to remove the purging of the sampling results. Just making a comment here as a reminder to discuss if we need to keep the purging as-is.

internal/keepers/service.go
```diff
@@ -364,28 +361,24 @@ func (wr *workerResults) FailureRate() float64 {
...
+	sur.blockKey = blockKey
+	sur.ok = true
 }
 
-func (sur *samplingUpkeepsResults) get() types.UpkeepResults {
+func (sur *samplingUpkeepsResults) get() (types.BlockKey, types.UpkeepResults, bool) {
 	sur.Lock()
-	results := make(types.UpkeepResults, len(sur.upkeepResults))
-	copy(results, sur.upkeepResults)
-	sur.upkeepResults = make(types.UpkeepResults, 0)
```
ferglor, line 387: We also weren't sure why the upkeep results got wiped after a get()?

internal/keepers/ocr.go
```diff
@@ -57,7 +65,26 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
...
+	if len(keys) > observationUpkeepLimit {
+		keys = keys[:observationUpkeepLimit]
+	}
+
+	obs := &ktypes.UpkeepObservation{
+		BlockKey:          blockKey,
+		UpkeepIdentifiers: []ktypes.UpkeepIdentifier{},
+	}
+
+	if len(keys) > 0 {
```
infiloop2, line 78: nit. do we need this if anymore?

internal/keepers/ocr.go
```diff
@@ -57,7 +65,26 @@ func (k *keepers) Observation(ctx context.Context, rt types.ReportTimestamp, _ t
 	// should be more uniform for all nodes
 	keys := keyList(filterUpkeeps(results, ktypes.Eligible))
 
-	b, err := limitedLengthEncode(keys, maxObservationLength)
+	// Check limit
+	if len(keys) > observationUpkeepLimit {
+		keys = keys[:observationUpkeepLimit]
+	}
```
infiloop2, line 71: nit. this might make more sense to do on `UpkeepIdentifiers` array after we've converted keys to upkeep identifiers

internal/keepers/ocr.go
```diff
@@ -92,9 +119,25 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+	// Build upkeep keys from the given observations
+	upkeepKeys, err := observationsToUpkeepKeys(attributed, k.reportBlockLag)
+	if err != nil {
+		return false, nil, fmt.Errorf("%w: failed to build upkeep keys from the given observations", err)
+	}
+
+	// Check the limit
+	if len(upkeepKeys) > reportKeysLimit {
+		upkeepKeys = upkeepKeys[:reportKeysLimit]
+	}
```
infiloop2, line 136: need to move this after shuffle and dedupe

internal/keepers/ocr.go
```diff
@@ -92,9 +119,25 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+
 	// pass the filter to the dedupe function
 	// ensure no locked keys come through
-	keys, err := shuffledDedupedKeyList(attributed, key, k.filter.Filter())
+	keys, err := shuffleUniqueObservations(upkeepKeys, key, k.filter.Filter())
```
infiloop2, line 140: nit. any reason to rename it? It's still doing the deduping i believe?

internal/keepers/service.go
```diff
@@ -178,37 +181,31 @@ func (s *onDemandUpkeepService) processLatestHead(ctx context.Context) {
 
 	// Get only the active upkeeps from the contract. This should not include
 	// any cancelled upkeeps.
-	keys, err := s.registry.GetActiveUpkeepKeys(ctx, "0")
+	blockKey, keys, err := s.registry.GetActiveUpkeepKeys(ctx, "0")
```
infiloop2, line 184: nit. rename to getLatestActiveUpkeepKeys?

internal/keepers/ocr.go
ferglor: We do not!

internal/keepers/ocr.go
ferglor: No I can change it back 👍

internal/keepers/service.go
ferglor: We need to keep the block key param because the implementation in pkg/chain uses it

internal/keepers/service.go
```diff
@@ -178,37 +181,31 @@ func (s *onDemandUpkeepService) processLatestHead(ctx context.Context) {
...
 	// select x upkeeps at random from set
 	keys = s.shuffler.Shuffle(keys)
-	size := s.ratio.OfInt(len(keys))
+	sampleSize := s.ratio.OfInt(len(keys))
 
-	s.logger.Printf("%d results selected by provided ratio %s", size, s.ratio)
-	if size <= 0 {
-		s.samplingResults.purge()
+	s.logger.Printf("%d results selected by provided ratio %s", sampleSize, s.ratio)
+	if sampleSize <= 0 {
```
infiloop2, line 197: change to `sampleSize < 0` as that is a true error condition. when sampleSize = 0, it will automatically flow through properly

internal/keepers/util.go
```diff
@@ -110,88 +111,146 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
+	var parseErrors int
+
+	res := make([][]ktypes.UpkeepKey, len(observations))
+
+	var allBlockKeys []ktypes.BlockKey
+	for i, observation := range observations {
 		// a single observation returning an error here can void all other
 		// good observations. ensure this loop continues on error, but collect
 		// them and throw an error if ALL observations fail at this point.
 		// TODO we can't rely on this concrete type for decoding/encoding
```
infiloop2, line 172: is this still valid?

internal/keepers/util.go
```diff
@@ -110,88 +111,146 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
-			ob = append(ob, o)
-		}
+		allBlockKeys = append(allBlockKeys, upkeepObservation.BlockKey)
 
-		sort.Sort(sortUpkeepKeys(ob))
-		kys[i] = ob
+		// if we have a non-empty list of upkeep identifiers, use the zeroth upkeep identifier
+		if len(upkeepObservation.UpkeepIdentifiers) > 0 {
+			res[i] = []ktypes.UpkeepKey{
+				chain.NewUpkeepKeyFromBlockAndID(upkeepObservation.BlockKey, upkeepObservation.UpkeepIdentifiers[0]),
```
infiloop2, line 185: this is hardcoded to be 0. instead we should pass `observationUpkeepsLimit` to this function as use that to limit slice

infiloop2: also not sure why we are appending block number here. we can just keep list of upkeep identifiers as we append the block number later anyway

internal/keepers/util.go
```diff
@@ -110,88 +111,146 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
 	}
 
-	if parseErrors == len(attributed) {
-		return nil, fmt.Errorf("%w: cannot prepare sorted key list; observations not properly encoded", err)
+	if parseErrors == len(observations) {
+		return nil, fmt.Errorf("%w: cannot prepare sorted key list; observations not properly encoded", ErrTooManyErrors)
 	}
 
-	keys, err := dedupe(kys, filters...)
+	// Here we calculate the median block that will be applied for all upkeep keys.
```
infiloop2, line 194: nit. more descriptive comment explaining role of reportBlockLag too.

internal/keepers/util.go
```diff
@@ -110,88 +111,146 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
-		}
+	return res, nil
+}
+
+func calculateMedianBlock(data []ktypes.BlockKey, reportBlockLag int) ktypes.BlockKey {
+	var blockKeyInts []*big.Int
+
+	for _, d := range data {
+		blockKeyInt := big.NewInt(0)
+		blockKeyInt, _ = blockKeyInt.SetString(string(d), 10)
```
infiloop2, line 228: what happens if it's not parseable as big int?

internal/keepers/ocr.go
```diff
@@ -92,13 +117,29 @@ func (k *keepers) Report(ctx context.Context, rt types.ReportTimestamp, _ types.
...
+	// Build upkeep keys from the given observations
+	upkeepKeys, err := observationsToUpkeepKeys(attributed, k.reportBlockLag)
+	if err != nil {
+		return false, nil, fmt.Errorf("%w: failed to build upkeep keys from the given observations", err)
+	}
+
 	// pass the filter to the dedupe function
 	// ensure no locked keys come through
-	keys, err := shuffledDedupedKeyList(attributed, key, k.filter.Filter())
+	keys, err := shuffleDedupedObservations(upkeepKeys, key, k.filter.Filter())
```
infiloop2, line 133: any reason to not do  `upkeepKeys, err := ...`
the variable `keys` creates confusion

internal/keepers/util.go
```diff
@@ -110,88 +111,146 @@ func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error)
...
-	var err error
+func filterDedupeShuffleObservations(upkeepKeys [][]ktypes.UpkeepKey, keyRandSource [16]byte, filters ...func(ktypes.UpkeepKey) bool) ([]ktypes.UpkeepKey, error) {
+	uniqueKeys, err := filterAndDedupe(upkeepKeys, filters...)
+	if err != nil {
+		return nil, err
+	}
 
-	if len(attributed) == 0 {
-		return nil, fmt.Errorf("%w: must provide at least 1", ErrNotEnoughInputs)
+	uniqueKeys, err = trimLowerBlocks(uniqueKeys)
```
infiloop2, line 120: do we really need this? not sure what this function is doing

internal/keepers/util.go
```diff
@@ -72,7 +73,7 @@ func (s sortUpkeepKeys) Len() int {
 	return len(s)
 }
 
-func dedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error) {
+func filterAndDedupe[T fmt.Stringer](inputs [][]T, filters ...func(T) bool) ([]T, error) {
 	if len(inputs) == 0 {
 		return nil, fmt.Errorf("%w: must provide at least 1", ErrNotEnoughInputs)
 	}
```
infiloop2, line 79: I'm not sure why this is needed. can this be removed?

pkg/chain/upkeep_observation.go
```diff
@@ -0,0 +1,10 @@
+package chain
+
+import (
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+)
+
+type UpkeepObservation struct {
```
infiloop2, line 7: I had to add this in chain as it needed a concrete type for BlockKey. Followed the same pattern as BlockKey and UpkeepKey.

I found this very confusing as to why we needed different interface and concrete type and the need to convert between each other in code. Also these types are in chain package which is confusing by itself, since some files here like evm registry is not actually used, but the concrete types are actually used in prod code. Would like to discuss how we can structure this better

EasterTheBunny: This is part of the design that needs to be revisited. I already have a working theory on how to do this so we can have that discussion whenever.

pkg/types/types.go
infiloop2: changed to 1,2

internal/keepers/ocr.go
infiloop2: resolved, using blockkey from head ticker

pkg/chain/evmregistry.go
```diff
@@ -47,8 +47,8 @@ func NewEVMRegistryV2_0(address common.Address, client types.EVMClient) (*evmReg
 	}, nil
 }
 
-func (r *evmRegistryv2_0) GetActiveUpkeepKeys(ctx context.Context, block types.BlockKey) ([]types.UpkeepKey, error) {
-	opts, err := r.buildCallOpts(ctx, block)
+func (r *evmRegistryv2_0) GetActiveUpkeepIDs(ctx context.Context) ([]types.UpkeepIdentifier, error) {
```
infiloop2, line 50: just to confirm @EasterTheBunny , this file is not actually used in prod right?

pkg/chain/upkeep_observation.go
infiloop2: also wasn't able to find an easy way other than copy the type in CL repo https://github.com/smartcontractkit/chainlink/pull/8457 . Definitely need to fix this before merging

infiloop2: created a todo for now, don't want to block this PR since this pattern is already present in codebase

infiloop2 (approved):

infiloop2 (dismissed):
lgtm. @EasterTheBunny could you also please review once before we merge, thanks!

begmaroman (dismissed):

internal/keepers/filter.go
```diff
@@ -121,19 +113,29 @@ func (rc *reportCoordinator) Accept(key types.UpkeepKey) error {
...
 
+	// Also apply an idBlock if the key is after an existing idBlock check key
 	blockKey, _, err := key.BlockKeyAndUpkeepID()
 	if err != nil {
 		return err
 	}
 
-	if bl, ok := rc.idBlocks.Get(string(id)); ok && string(bl.KeyBlockNumber) > string(blockKey) {
-		return nil
+	bl, ok := rc.idBlocks.Get(string(id))
```
EasterTheBunny, line 125: Doing `string(id)` here still invites ids like `00001` into the comparison because the input is not getting validated anywhere. We have explicit json parsing implementations for keys, but the observation was changed to an array of identifiers. 

This needs to follow the same paradigm as `UpkeepKey` where the type or interface implements `fmt.Stringer` AND it needs explicit parsing and validation functions.

EasterTheBunny (changes requested):

internal/keepers/filter.go
infiloop2: The core issue we need to check is validation of observations which we get from nodes. I'll fix that in this PR. 
Our overall types (concrete and interfaces) and getting really messed up and I am not sure that's the correct way to go about converting all types to interfaces. Will do a thorough cleanup of that in a separate PR

EasterTheBunny: The observation validation looks good. Thanks!

EasterTheBunny (approved):

PR #73, opened 2023-02-14
Close of Closed Channel in Worker Group

EasterTheBunny:
The `Stop()` method closes a channel, which causes all workers in the worker group to exit. The finalizer also attempts to close the same channel.

The finalizer always runs during a garbage collection cycle when no more references exist to the WorkerGroup. If the `Stop()` function is called first, the finalizer will attempt to close a closed channel.

The solution is to wrap the contents of the `Stop()` function in a sync.Once and call the stop function from the finalizer instead of closing the channel directly.


ferglor (approved):
Cool! Thanks for the explanation, I really appreciate it! 🙌

infiloop2:
woops, screwed up rebasing when trying to merge. To fix, I just cherry picked the commit with the fix and pushed it directly to `feature/auto-2015-observation-type`. Closing this PR. 

Thanks for fixing this @EasterTheBunny !

PR #74, opened 2023-02-15
Nancy Dependency Scan in CI Pipeline

EasterTheBunny:
The change here adds Nancy scans to the CI pipeline just after unit tests to ensure dependencies are scanned for vulnerabilities on a regular basis.


go.mod
```diff
@@ -2,43 +2,48 @@ module github.com/smartcontractkit/ocr2keepers
...
+	github.com/go-echarts/go-echarts/v2 v2.2.5
+	github.com/google/uuid v1.3.0
+	github.com/pkg/errors v0.9.1
 	github.com/shopspring/decimal v1.3.1
-	github.com/smartcontractkit/libocr v0.0.0-20221209172631-568a30f68407
-	github.com/spf13/pflag v1.0.3
-	github.com/stretchr/testify v1.7.2
-	go.uber.org/multierr v1.6.0
-	golang.org/x/crypto v0.0.0-20221012134737-56aed061732a
+	github.com/smartcontractkit/libocr v0.0.0-20230112103111-f100435787c8
```
infiloop2, line 17: what is the version used in CL repo?

.nancy-ignore
```diff
@@ -0,0 +1,16 @@
+# The go-libp2p dependency is not used directly in this plugin so the following
```
infiloop2, line 1: i'm not sure if we're even using libp2p for ocr anymore. is this copied from CL repo?

go.mod
EasterTheBunny: The chainlink node uses commit 568a30f but this repo uses the latest commit of f100435.

.nancy-ignore
EasterTheBunny: No. I made this comment to indicate that this repo only really uses libocr for types and interfaces and doesn't use libp2p directly. The libocr repo may still have this as a dependency even if the chainlink repo uses a different implementation. OCR2 allows for an arbitrary networking layer as it defines an expected interface, not a distinct type.

go.mod
EasterTheBunny: So not use the latest?

infiloop2: yep

EasterTheBunny: done

infiloop2 (approved):

PR #75, opened 2023-02-15
removed unnecessary or outdated todo items

EasterTheBunny:


internal/keepers/ocr.go
```diff
@@ -225,7 +225,6 @@ func (k *keepers) ShouldTransmitAcceptedReport(_ context.Context, rt types.Repor
 		transmitConfirmed := k.filter.IsTransmissionConfirmed(id.Key)
 		// multiple keys can be in a single report. if one has a confirmed transmission
 		// (while others may not have), don't try to transmit again
-		// TODO: reevaluate this assumption
```
infiloop2, line 228: not sure about this. I checked functions design, and they transmit if atleast one of the requests has not been confirmed, since there's value in that tx. wdyt?

EasterTheBunny: Right. The confirmation is by key instead of by report, but each key will still send a perform log even if it failed. So if any key in a report is confirmed by a node, all keys in the same report should be confirmed as well.

infiloop2: not necessarily. An upkeep can be pre-filtered out in the contract. I don't think this will quite happen under normal circumstances, especially as we have batch size = 1 to begin with. But i do not want to remove this todo as this requires more thought

EasterTheBunny: Ticket created to capture this TODO.

infiloop2 (approved):

PR #76, opened 2023-02-15
Change the BlockKey type to an interface

ferglor:


pkg/chain/block_key.go
```diff
@@ -0,0 +1,32 @@
...
+	b, ok = b.SetString(kk.String(), 10)
+	if !ok {
+		return false, ErrBlockKeyNotParsable
+	}
+
+	if gt := a.Cmp(b); gt > 0 {
+		return true, nil
+	}
+	return false, nil
+}
```
begmaroman, line 28:
```go
func (k BlockKey) After(kk types.BlockKey) (bool, error) {
	a, ok := big.NewInt(0).SetString(k.String(), 10)
	if !ok {
		return false, ErrBlockKeyNotParsable
	}

	b, ok = big.NewInt(0).SetString(kk.String(), 10)
	if !ok {
		return false, ErrBlockKeyNotParsable
	}

	gt := a.Cmp(b)
	return gt > 0, nil
}
```
wdyt?

begmaroman (approved):

internal/keepers/filter.go
```diff
@@ -87,17 +87,17 @@ func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
...
 			}
 
 			// Return false if empty
-			if bl.TransmitBlockNumber == "" {
+			if bl.TransmitBlockNumber == nil || bl.TransmitBlockNumber.String() == "" {
 				return false
 			}
 
-			transmitBlockNumber, err := strconv.Atoi(string(bl.TransmitBlockNumber))
+			transmitBlockNumber, err := strconv.Atoi(bl.TransmitBlockNumber.String())
```
infiloop2, line 100: in line 106 below, can we use after?

PR #77, opened 2023-02-20
Handle stale report logs

infiloop2:
Corresponding CL PR https://github.com/smartcontractkit/chainlink/pull/8526


internal/keepers/filter.go
```diff
@@ -165,37 +152,112 @@ func (rc *reportCoordinator) checkLogs() {
...
 	}
 }
 
+// This function tries to update idBlock for a given key to val. If no idBlock existed for this key
+// then it's just updated with val, however if it existed before then it is only updated if checkBlockNumber
+// is set higher, or checkBlockNumber is the same and transmitBlockNumber is higher.
+//
+// For a sequence of updates, updateIdBlock can be called in any order on different nodes, but by
+// maintaining this invariant it results in an eventually consistent value across nodes.
+func (rc *reportCoordinator) updateIdBlock(key string, val idBlocker) {
```
infiloop2, line 202: need some suggestions to make this function more elegant

internal/keepers/ocr.go
```diff
@@ -235,14 +235,6 @@ func (k *keepers) ShouldAcceptFinalizedReport(_ context.Context, rt types.Report
...
 	}
 
-	for _, r := range results {
-		// check whether the key is already accepted
-		if k.filter.CheckAlreadyAccepted(r.Key) {
-			k.logger.Printf("%s: key already accepted: %s", r.Key, lCtx.Short())
-			return false, fmt.Errorf("failed to accept key %s as it was already accepted", r.Key)
-		}
-	}
-
```
infiloop2, line 245: removing this error to have consistent behaviour that we accept all reports we get. If we get it for the same checkBlockNumber it might not result in any update to the filters, but throwing an error can cause issues if there were other upkeeps in batch

internal/keepers/filter.go
```diff
@@ -159,39 +149,76 @@ func (rc *reportCoordinator) checkLogs() {
...
 
+			bl, err := l.TransmitBlock.Subtract(rc.reportBlockLag)
+			if err != nil {
+				rc.logger.Printf("Error while subtracting reportBlockLag from TransmitBlock")
+				continue
+			}
+			rc.updateIdBlock(string(l.UpkeepId), idBlocker{
+				CheckBlockNumber: l.TransmitBlock, // As we do not have the actual checkBlockNumber which generated this
+				//reorg log, use transmitBlockNumber to override all previous checkBlockNumbers
+				TransmitBlockNumber: chain.BlockKey(bl), // Removes the id from filters from l.TrnasmitBlock - reportBlockLag.
```
infiloop2, line 219: TODO: can we set to 0?

internal/keepers/ocr.go
EasterTheBunny: So this change was because `Accept` avoids changing the filters now? We put this in for the reason of exiting early on report generation if at least one key had already been accepted.

EasterTheBunny (changes requested):

internal/keepers/ocr.go
infiloop2: I am not sure if we had a strong reason for it to exist in first place. For reports containing multiple upkeeps, if one key was accepted before and one wasn't then it could cause issues, some nodes may accept the report some might not

Having a uniform logic that we accept all reports helps avoid edge cases when a node's local filters can get out of sync

internal/keepers/filter.go
```diff
@@ -165,33 +149,89 @@ func (rc *reportCoordinator) checkLogs() {
...
+
+				rc.updateIdBlock(string(id), idBlocker{
+					CheckBlockNumber:    logCheckBlockKey,
+					TransmitBlockNumber: l.TransmitBlock,
+				})
 			}
+		}
+	}
+
+	slateReportLogs, _ := rc.logs.StaleReportLogs(context.Background())
```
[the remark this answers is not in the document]
infiloop2: fixed

internal/keepers/filter.go
```diff
@@ -165,33 +149,89 @@ func (rc *reportCoordinator) checkLogs() {
...
-					rc.logger.Printf("Higher check block already exists in idBlocks, not clearing idBlocks while processing perform log for key %s", l.Key)
-					continue
-				}
+			rc.updateIdBlock(string(id), idBlocker{
+				CheckBlockNumber:    logCheckBlockKey,
+				TransmitBlockNumber: l.TransmitBlock, // Removes the id from filters from higher blocks
+			})
+		}
+
+		if ok && confirmed {
```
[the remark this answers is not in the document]
infiloop2: fixed

internal/keepers/ocr.go
EasterTheBunny: I think the problem is that when `Accept` returns an error, we exit building the report. However, since `Accept` changes the filter state, exiting without a report blocks keys from being added to the next report. Keys will end up getting blocked, but not from having a report generated. We won't see the effects of this with one key per report, but it will come back once we start doing multiple per report again.

EasterTheBunny: The consequence of that will be more logs with upkeep perform errors, right? What if all keys return an error in `Accept`? We should avoid sending a report at that point.

infiloop2: updated should transmit logic per offline discussion https://github.com/smartcontractkit/ocr2keepers/pull/77/commits/7df09b43c6d6e1a7109b0f57c1242e06c63faee8

EasterTheBunny (approved):

PR #82, opened 2023-02-22
AUTO-2152: Upkeep id manipulation

ferglor:
In this PR, I've added the following test cases:

- Node sending negative number as upkeep id
- Node sending 0 upkeep id
- Node sending upkeep id with 1000 or more digits
- Node sending multiple upkeep IDs with leading zeroes and different block numbers
- Node sending multiple upkeep IDs with leading zeroes and same block numbers

### DIFF
--- a/internal/malicious/blocks.go
+++ b/internal/malicious/blocks.go
@@ -256,50 +256,12 @@ func SendLeadingZeroUpkeepIDs(ctx context.Context, original []byte, _ error) (st
 	}
 
 	var ids []types.UpkeepIdentifier
-	for i := 0; i < len(ob.UpkeepIdentifiers); i++ {
-		keyStr, err := GenerateRandomASCIIString(1000)
-		if err != nil {
-			return name, nil, err
-		}
-
-		ids = append(ids, types.UpkeepIdentifier(fmt.Sprintf("%s%s", strings.Repeat("0", rand2.Intn(100)), keyStr)))
+	for _, upkeepID := range ob.UpkeepIdentifiers {
+		ids = append(ids, types.UpkeepIdentifier(fmt.Sprintf("%s%s", strings.Repeat("0", rand2.Intn(100)), upkeepID)))
 	}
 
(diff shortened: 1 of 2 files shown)


internal/malicious/blocks.go
```diff
@@ -145,6 +147,166 @@ func SendVeryLargeBlockValue(ctx context.Context, original []byte, _ error) (str
...
+
+	var ids []types.UpkeepIdentifier
+	for i := 0; i < len(ob.UpkeepIdentifiers); i++ {
+		keyStr, err := GenerateRandomASCIIString(1000)
+		if err != nil {
+			return name, nil, err
+		}
+
+		ids = append(ids, types.UpkeepIdentifier(fmt.Sprintf("%s%s", strings.Repeat("0", rand2.Intn(100)), keyStr)))
+	}
```
EasterTheBunny, line 266: I think the idea I originally had for this variant was to use the ids already in the observation and prepend zeros to them. A test above is already testing the random ascii strings, so no need to include it again here.

internal/malicious/blocks.go
```diff
@@ -145,6 +147,166 @@ func SendVeryLargeBlockValue(ctx context.Context, original []byte, _ error) (str
...
+
+	b, err := json.Marshal(observation)
+	return name, b, err
+}
+
+// SendLeadingZeroUpkeepIDsSameBlock produces an encoded json object with upkeep IDs with leading zeroes for the same block key
+func SendLeadingZeroUpkeepIDsSameBlock(ctx context.Context, original []byte, _ error) (string, []byte, error) {
+	name := "Leading Zero Upkeep IDs With Same Block"
+
```
EasterTheBunny, line 280: I think this test case is probably superfluous now that the block key is separate from upkeep ids. Originally, it was upkeep keys and the idea was to have different block number per key.

EasterTheBunny (changes requested):

EasterTheBunny (approved):

PR #83, opened 2023-02-23
Improve validation against negative keys

infiloop2:
Only allow non-negative numbers in observations

### DIFF
--- a/internal/keepers/util_test.go
+++ b/internal/keepers/util_test.go
@@ -180,6 +180,20 @@ func Test_observationsToUpkeepKeysError(t *testing.T) {
 				ktypes.UpkeepIdentifier("03"), // Error in both
 			},
 		},
+		{
+			BlockKey: chain.BlockKey("-1"), // Should be non negative
+			UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+				ktypes.UpkeepIdentifier("2"),
+				ktypes.UpkeepIdentifier("3"),
+			},
+		},
+		{
+			BlockKey: chain.BlockKey("1"), // Should be non negative
+			UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+				ktypes.UpkeepIdentifier("-2"),
+				ktypes.UpkeepIdentifier("3"),
+			},
+		},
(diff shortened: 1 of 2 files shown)


ferglor (dismissed):

ferglor:
Looks good, are there any test cases we could add to these new validations?

internal/keepers/util_test.go
```diff
@@ -180,6 +180,20 @@ func Test_observationsToUpkeepKeysError(t *testing.T) {
...
+			BlockKey: chain.BlockKey("-1"), // Should be non negative
+			UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+				ktypes.UpkeepIdentifier("2"),
+				ktypes.UpkeepIdentifier("3"),
+			},
+		},
+		{
+			BlockKey: chain.BlockKey("1"), // Should be non negative
+			UpkeepIdentifiers: []ktypes.UpkeepIdentifier{
+				ktypes.UpkeepIdentifier("-2"),
```
ferglor, line 193: Should this line should be commented to state that the UpkeepIdentifier should be non-negative?

ferglor (approved):
🎉

pkg/chain/upkeep_observation.go
```diff
@@ -53,6 +56,9 @@ func (u *UpkeepObservation) validate(uo upkeepObservation) error {
 		if uiInt.String() != string(ui) {
 			return ErrInvalidUpkeepIdentifier
 		}
+		if uiInt.Cmp(big.NewInt(0)) == -1 {
```
EasterTheBunny, line 59: Should this be `<= 0` instead since zero value identifiers are also invalid?

infiloop2: I considered it, technically upkeep ID is uint256 so it can be 0. I'll add it for the blockKey though

ferglor (approved):

pkg/chain/upkeep_observation.go
EasterTheBunny: Ahh. ok. Sounds good.

PR #88, opened 2023-03-01
description of structural changes

EasterTheBunny:


infiloop2 (changes requested):
This has gone way beyond the original scope of refactor. We need clear motivation and list of problems to solve before we dig more into alternate designs

EasterTheBunny:
Switched to new branch focused on refactor only. The re-architecture of the protocol is tabled for now.

PR #90, opened 2023-03-08
added document that describes in high level how functions work

EasterTheBunny:


OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+
+An automation OCR2 observation is constructed of a block number and a shuffled
+list of upkeep IDs (big.Int), json encoded to a limit of max bytes. The list
+should only include eligible upkeeps (covered in a later section). Shuffling
+is done using a randomness source derived from the config digest, epoch, and
+block number. This ensure all nodes have the same source, but the results are
+random across different OCR rounds if an observation is repeated.
+
+> the shuffling should not be necessary in the observation because we 
+> cannot trust that shuffling was done once the observation is consumed in Report
```
infiloop2, line 17: We still expect good nodes to shuffle their observations in order to cycle over the upkeeps so it is needed

OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+    return (observationBytes)
+```
+
+## How a Report is Constructed
+
+Observations from all other nodes (including the one processing a report) are
+processed into a final encoded report ready to be transmitted on chain. Key
+checks include:
+
+- all observations must pass decoding
```
infiloop2, line 59: some can fail here

OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+- a report must be built off of at least 1 observation
+- all observations must have a block number to calculate median from
+- observation id length should be less than or equal to max
+- check keys should be less than max keys for a report
```
infiloop2, line 63: what are the check keys here? Upkeep IDs?

OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+- all upkeeps should be checked at specified median block for eligibility
+- report capacity should be based on total gas used and max upkeeps per report
```
infiloop2, line 65: Should mention the report capacity is limited to 1 and is not affected by upkeep gas limit

OVERVIEW.md
EasterTheBunny: This is because we're limiting the observation size?

infiloop2: yes

OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+## How a Report is Constructed
+
+Observations from all other nodes (including the one processing a report) are
+processed into a final encoded report ready to be transmitted on chain. Key
+checks include:
+
+- all observations must pass decoding
+- a report must be built off of at least 1 observation
```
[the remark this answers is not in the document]
EasterTheBunny: This is saying that we need to see an eligible upkeep from at least 1 node to start the report process. The f+1 is only on matching reports, not observations.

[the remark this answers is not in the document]
infiloop2: f+1 is the minimum number of good observations we will always have when report function is called

OVERVIEW.md
EasterTheBunny: Right.

OVERVIEW.md
EasterTheBunny: fixed

OVERVIEW.md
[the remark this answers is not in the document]
EasterTheBunny: f+1 is the minimum number of matching reports, not observations. we don't do quorum checks on incoming observations. it takes only 1 observation from 1 node to trigger all nodes to verify that observation and build a corresponding report.

OVERVIEW.md
EasterTheBunny: updated with more explanation

OVERVIEW.md
```diff
@@ -0,0 +1,483 @@
...
+
+## How a Report is Accepted
+
+In the acceptance of a report, the details of the incoming report must match the
+following:
+
+- byte length should be greater than zero
+- at least one upkeep in report
+- no decode error
+- no errors from marking a key as 'in-flight'
```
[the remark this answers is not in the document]
EasterTheBunny: updated with more explanation

OVERVIEW.md
EasterTheBunny: updated with more information

infiloop2 (approved):

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+define [blockNumber]
+define [upkeepResults]
+for each (upkeepResults)
+    if [upkeepResult] is Pending
+        remove (upkeepResult) from (upkeepResults)
+
+return (blockNumber), (upkeepResults), and no error
+```
+
+### Filter
```
coventry, line 304: This section is obviously quite important, but not very clearly written, at least from my perspective. I suggest careful review and reformulation of the design decisions here, as errors and oversights may be hiding in the unclear parts.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+for each (performLogs)
+    define [upkeepKey]
+    define [performLogConfirmations]
+    define [performBlockNumber]
+
+    if (performLogConfirmations) >= (configMinConfirmations)
+        split (upkeepKey) into [blockNumber], [upkeepID]
+
+        if (upkeepKey) in (pendingKeys)
+            define [upkeepKeyConfirmationState]
```
coventry, line 421: Not clear what's going on here.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+
+                    set (upkeepID) in (blockedIDs) with (blockNumber) and (performBlockNumber)
+
+```
+
+It can happen that in between the time the report is generated and it gets
+confirmed on chain something changes and it becomes stale. Current scenarios are:
+
+- Another report for the upkeep is transmitted making this report stale
+- Reorg happens which changes the checkBlockHash making this reorged report as it was checked on a different chain
```
coventry, line 440: I don't have a sufficiently precise understanding of what's going on here to assess whether a re-org can put the system in a bad state.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+- There's a massive gas spike and upkeep does not have sufficient funds when report gets on chain
+
+In such cases the upkeep is not performed and the contract emits a log
```
coventry, line 443: The contract is not aware of re-orgs?

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+indicating the staleness reason instead of UpkeepPerformed log. We don't have 
+different behaviours for different staleness reasons and just want to unlock the
+upkeep when we receive such log.
+
+For these logs we do not have the exact key which generated this log. Hence we
```
coventry, line 448: This sounds important, but it's not clear to me what it's saying.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+define [configMinConfirmations]
+define [pendingKeys]
+define [blockedIDs]
+
+for each (performLogs)
+    define [upkeepKey]
+    define [performLogConfirmations]
+    define [performBlockNumber]
+
+    if (performLogConfirmations) >= (configMinConfirmations)
```
coventry, line 464: Not clear what's going on, here.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+This limits the workload of individual nodes and attempts to distribute the 
+total workload equally to all nodes while synchronizing only on `n`, `p`, 
+and `r`.
+
+Faulty node limit `f` is included in the calculation of the sampling ratio.
+
+**Polling**
+
+For each block, this service attempts to check the state of all active upkeeps
+provided by the registry. First, all upkeeps are shuffled before checking
```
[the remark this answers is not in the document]
coventry: I talked this over with Lorenz, and he pointed out that I'm ignoring the need to run the check-upkeep method, which inherently requires an RPC call.

OVERVIEW.md
```diff
@@ -0,0 +1,484 @@
...
+## How a Report is Constructed
+
+Observations from all other nodes (including the one processing a report) are
+processed into a final encoded report ready to be transmitted on chain. Key
+checks include:
+
+- all observations that don't pass decoding are discarded
+- a report must be built off of at least 1 observation
+- all observations must have a block number to calculate median from
+- observation id length should be less than or equal to max
```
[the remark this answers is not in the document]
infiloop2: TODO: look into protobuf encoding

PR #91, opened 2023-03-08
Move dependency checks to schedule

EasterTheBunny:
Dependency checks can be run on a schedule instead of as part of pull requests or pushes that block the pipeline. These will only be notifications so they should be monitored.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -37,22 +37,6 @@ jobs:
     - name: Run Unit Tests
       run: go test -v ./...
 
-  dependency-check:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v3
-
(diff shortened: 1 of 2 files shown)


EasterTheBunny:
The CL repo doesn't to dependency checks on this repo, so they're still needed here. This change unblocks the pipeline and runs the checks on a schedule instead. It makes issues informational instead of blockers.

infiloop2:
My question is more about why have this separate repo at all and manage our own dependency process as opposed to moving packages from here to CL repo

EasterTheBunny:
While that may still be in debate, is it a blocker to make this change?

infiloop2:
Not blocking this change, but further investment into this

infiloop2 (approved):

PR #94, opened 2023-03-15
Plugin Refactor Primary PR

EasterTheBunny:
This change set attempts to define high-level concepts on refactoring the plugin. Concrete types include a versioned observation as well as byte arrays for blocks, identifiers, and keys.


docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
...
+initial intent is to not make a breaking change to the existing registry
+implementation by making a major interface change. The only change here is
+to return the above described type which will just be a wrapper for the existing
+type. An observer does not need to know any of the details of the result. It 
+only needs to be able to pass the result along to an `Encoder`.
+
+```
+type Registry interface {
+    CheckUpkeep(context.Context, ...UpkeepKey) ([]UpkeepResult, error)
+}
```
RyanRHall, line 38: would we want to call this a `Registry` for observation types like cron and log triggers? we're not actually talking to the registry contract for those trigger types.

docs/400_REPORT.md
```diff
@@ -0,0 +1,72 @@
...
+In the context of OCR Automation, a report contains all data required to perform
+one or more upkeeps per user specifications including all calldata.
+
+## Building a Report
+All nodes receive some set of observations broadcast from other nodes. A report
+is generated by finding the intersection of all observations such that:
+
+- at the intersection, the number of observations meets or exceeds the quorum
+- each intersection point should be a unique upkeep
+- if intersections duplicate an upkeep, the most recent `PointState` is used
```
RyanRHall, line 11: is "most recent" manipulable by a malicious node?

docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
+# What is a registry?
+
+A `Registry` should be responsible for contract specific behavior and consolidate
```
cmalec, line 3: do we want to call this registry? I feel like it makes it a little confusing since it doesn't seem to be actually representing the Registry contract.

docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
...
+required to determine if an upkeep is eligible to be performed or not. The
+initial intent is to not make a breaking change to the existing registry
+implementation by making a major interface change. The only change here is
+to return the above described type which will just be a wrapper for the existing
+type. An observer does not need to know any of the details of the result. It 
+only needs to be able to pass the result along to an `Encoder`.
+
+```
+type Registry interface {
+    CheckUpkeep(context.Context, ...UpkeepKey) ([]UpkeepResult, error)
```
cmalec, line 37: Like here `Registry.CheckUpkeep` is not actually mapping to the Registry contracts function checkUpkeep right? This is just like the idea of simulating the flow for eligibility?

docs/500_ENCODER.md
```diff
@@ -0,0 +1,50 @@
...
+    EncodeReport([]UpkeepResult, ...Config) ([]byte, error)
+    // EncodeUpkeepIdentifier should take the output of a registry check result
+    // as input and return an UpkeepIdentifier
+    EncodeUpkeepIdentifier(UpkeepResult) (UpkeepIdentifier, error)
+    // KeysFromReport extracts all upkeep keys from a report byte array
+    KeysFromReport([]byte) ([]UpkeepKey, error)
+    // Eligible takes an upkeep result to provide an abstraction per registry;
+    // returns whether an upkeep is eligible or not along with any decoding
+    // or type conversion errors encountered
+    Eligible(UpkeepResult) (bool, error)
```
cmalec, line 48: just so im sure im on the same page this is just checking the UpkeepResult struct to determine if its Eligible its not doing any calls to get data from anywhere else.

docs/diagrams/POLLINGOBSERVER.md
```diff
@@ -0,0 +1,44 @@
...
+
+> PollingObserver implements the `Observer` interface, but details of that
+> interface are not shown here.
+>
+> A potential attack vector is flooding a registry with eligible upkeeps. As a
+> mitigation, limit the number of eligible upkeeps per head and randomly select
+> them from the total set of eligible before adding them to the cache.
+
+```mermaid
+sequenceDiagram
```
cmalec, line 11: So this is just determining what upkeeps are in the observation group. No eligibility is decided here

docs/diagrams/PLUGINDIAGRAM.md
```diff
@@ -0,0 +1,100 @@
...
+
+        loop each upkeep
+            Observer->>Coordinator: IsPending()
+            Coordinator-->>Observer: bool
+            
+            Observer->>Observer: addToObservation
+
+            Observer->>Encoder: Eligible()
+            Encoder-->>Observer: bool
+        end
```
cmalec, line 28: Here is where the observer is constantly checking eligibility correct? The encoder is not really doing any work its just the translator for the observer as the observer doesnt want to interpret the UpkeepResult itself?

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+    // compatibility to previous versions
+    Data       map[int]PointState
+}
+```
+
+#### Report
+A report should be able to be constructed without ever checking internal or
+external state. All observations should be fully validated independently, merged 
+together, and reduced by provided pending state structs. The result should be 
+entirely stateless.
```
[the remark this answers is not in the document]
EasterTheBunny: This is ideal and not for the refactor. :( But maybe it can be coming in a later version.

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
...
+An `Observer` evaluates upkeep state from a chain registry and collects upkeeps
+that need to be performed along with data required for comparison and reporting.
+
+Examples can include:
+- polling observer (polls the chain at each block)
+- event trigger observer (logs can trigger observations)
+- scheduled observer (polling occurs on a defined schedule like cron)
+
+## Observer Rules
+- an observer should only be used for observation triggers and timing
```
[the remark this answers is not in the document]
EasterTheBunny: A log trigger should be an observer type. Think of it as what triggers polling and how polling is scoped. The current version is triggered by a new block, selects all upkeeps for polling, and starts again at the next block perpetually.

Instead, a log can trigger polling, WHAT is polled is contained in the trigger, and how long polling occurs (if no report is ever generated) can be a configuration. Let's say a log triggers the observer to start polling on upkeep `x` for `y` blocks or until a report is detected. In the end, it's a modified observer.

docs/350_REGISTRY.md
EasterTheBunny: The `Registry` type is intended to be an intermediate layer between an Observer and the contract. The idea is that it can offer an abstraction over contract specific actions. In this case is `CheckUpkeep` where the contract will perform `checkUpkeep` and `performUpkeep`. In off-chain lookup, there will also be other actions too, all wrapped up in the `CheckUpkeep` function.

Also there should be room for other functionality that might be needed for log and cron base observers such as per-upkeep configurations.

docs/400_REPORT.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Accepting a Report
+A report should be accepted if all upkeeps it identifies in the report are not
+pending. Once the report passes this test, all upkeeps should be marked as
+pending to all observers.
+
+## Transmitting a Report
+A report should be transmitted if all the upkeeps it identifies in the report
+are not pending or the report identifier is unlocked. Different version will
+need to be supported. For the former,
```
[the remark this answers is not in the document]
EasterTheBunny: This is also for the re-architecture. Disregard for now.

docs/350_REGISTRY.md
EasterTheBunny: I'm open to another name. I agree with the confusion and it's just supposed to represent higher level ideas that one would want the contract to do. Here, the idea of checking the performability of an upkeep continues to change at the contract level, but the concept / output is still the same.

docs/350_REGISTRY.md
EasterTheBunny: Correct. The output `Result` should be everything necessary to build a report AND to determine if the upkeep is eligible to be performed.

docs/500_ENCODER.md
EasterTheBunny: Correct. Because each contract is likely to have structural differences of what exactly is in an `UpkeepResult`, this allows those changes to be non-breaking for the plugin where an `Encoder` is a translator for the contract.

docs/diagrams/PLUGINDIAGRAM.md
EasterTheBunny: Correct. This helps abstract the idea of a polling observer from any specific contract. All we need is a single observer type (polling, log trigger, cron, etc.) that can pair with any number of contracts provided the concept of eligible remains the same and an encoder is paired with the observer.

docs/100_OVERVIEW.md
FelixFan1992, line 1: some of the links are broken. can you fix them?

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+More about specifics of an observation from an observer here:
+[OBSERVATION](./300_OBSERVATION.md)
+
+##### Observation Struct
+An observation is composed of a list of eligible upkeeps and a list of pending
+upkeeps. Providing both allows for all nodes to produce a report without
+checking internal state. Versioning allows for version checks to be applied as
+new nodes can join the network with different active versions.
+
+```
```
FelixFan1992, line 50: how are these structs related to the `types.Observation` from `"github.com/smartcontractkit/libocr/offchainreporting2/types"`?

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+    // OtherStuff in this case is addative and the version minor is incremented
+    OtherStuff int
+}
+
+// ObservationV20 displays a potential way to differentiate observation versions
+type ObservationV20 struct {
+    VersionedObservation
+    // Data in this version is a different structure and would break 
+    // compatibility to previous versions
+    Data       map[int]PointState
```
FelixFan1992, line 84: curious why you envision a map of point states here? this is only a single observation so cannot compute a point state yet?

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+final transport type.
+
+More about specifics of an observation from an observer here:
+[OBSERVATION](./300_OBSERVATION.md)
+
+##### Observation Struct
+An observation is composed of a list of eligible upkeeps and a list of pending
+upkeeps. Providing both allows for all nodes to produce a report without
+checking internal state. Versioning allows for version checks to be applied as
+new nodes can join the network with different active versions.
```
FelixFan1992, line 48: do you see there could be more than 1 version of observations in a DON? maybe it's like we release a new version core node which has a new ocr2 keeper plugin. and external Node Ops upgrade at different times?

i feel this will make Report function and any other observation bytes decoding process more difficult.

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+
+The intended approach for extracting data from observations involves decoding
+the concrete plugin observation type, `VersionedObservation`, and further
+extracting values from the underlying observation using the appropriate encoder.
+
+This allows concrete types to be used directly by the plugin, where observation
+comparison is concerned, while leaving encoding a separate and modular 
+responsibility.
+
+Observation distinct types and decoders should be exposed as public types and
```
FelixFan1992, line 137: nice :)

docs/100_OVERVIEW.md
EasterTheBunny: A `types.Observation` is just a wrapper for `[]byte`, but the `VersionedObservation` provides structure and change management as the plugin evolves.

docs/100_OVERVIEW.md
EasterTheBunny: This is a future concept. A `PointState` is part of a re-architecture, so that can be ignored for now. The premise here is that a VersionedObservation can have different data structures per version.

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
+# OCR2 Automation Design Overview
+The intent of the plugin design is to separate responsibility of components,
+ensure modularity, and provide accessible types for viewing content in 
+messages via external sources (like telemetry).
+
+## Glossary of Terms
+- Observation: two different types; one from the `Plugin` and one from an `Observer`
+- Report: encoded instructions and/or values for on-chain execution
+- BlockKey: a block identifier (typically a big.Int on many chains)
```
infiloop2, line 9: nit. for now we can safely assume it will be uint64 on all chains until a new case comes up

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+- UpkeepIdentifier: an identifier for an upkeep (typically a big.Int)
```
infiloop2, line 10: nit. similar we can safely hardcode the assumption it will be uint256

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+- NetworkConfig: also called off-chain config in OCR; network configuration
+- In-flight: upkeeps are in-flight if they are reported, but not transmitted
+- Transmit: an attempt to include a report on-chain
+
+## Components
+
+### Plugin
+The `Plugin` provides methods for LibOCR to call as a valid OCR2 plugin. The 
+responsibility of each method is defined below. Generally, a `Plugin` gets
+observations from a list of multiple `Observer`, uses a `Merger` to merge
```
infiloop2, line 21: yet to read it completely, but q at this point is the need for multiple observers instead of a single one

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+observations from the network, and an `Encoder` to produce a transmittable
+report.
+
+#### Rules
+- A `Plugin` only interacts with `Observer`, `Merger`, and `Encoder`
+- A `Plugin` should not know anything about the structural contents of a `BlockKey`
+- A `Plugin` should not know anything about the structural contents of a `UpkeepIdentifier`
```
infiloop2, line 28: not sure if we need these. they can be safely hardcoded to uint64 and uint256

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+function should never directly interact with the encoded upkeep results. Its 
+sole responsibilities should remain as a collector and basic encoder to the 
+final transport type.
+
+More about specifics of an observation from an observer here:
+[OBSERVATION](./300_OBSERVATION.md)
+
+##### Observation Struct
+An observation is composed of a list of eligible upkeeps and a list of pending
+upkeeps. Providing both allows for all nodes to produce a report without
```
infiloop2, line 46: I would split concerns of adding pending upkeeps to this refactor. There are additional complexities there (blocked till particular block number, and number of pending upkeeps) that are best explored in a separate discussion

docs/100_OVERVIEW.md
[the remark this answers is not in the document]
infiloop2: Just to clarify are you envisioning a need for report to merge different observation versions here?

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
+# Generic Observer
+An `Observer` evaluates upkeep state from a chain registry and collects upkeeps
+that need to be performed along with data required for comparison and reporting.
+
+Examples can include:
+- polling observer (polls the chain at each block)
+- event trigger observer (logs can trigger observations)
+- scheduled observer (polling occurs on a defined schedule like cron)
```
infiloop2, line 8: Is it correct to say that observer can be of different types and can be triggered differently but the plugin will always poll observer in each round?

docs/100_OVERVIEW.md
```diff
@@ -0,0 +1,147 @@
...
+observations from a list of multiple `Observer`, uses a `Merger` to merge
+observations from the network, and an `Encoder` to produce a transmittable
+report.
+
+#### Rules
+- A `Plugin` only interacts with `Observer`, `Merger`, and `Encoder`
+- A `Plugin` should not know anything about the structural contents of a `BlockKey`
+- A `Plugin` should not know anything about the structural contents of a `UpkeepIdentifier`
+- A `Plugin` should start and stop services per its appropriate life cycle
+- A `Plugin` should treat all incoming observations as potentially malicious
```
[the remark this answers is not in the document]
EasterTheBunny: Right. In the report function. I can expand on that for clarity.

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Observer Rules
+- an observer should only be used for observation triggers and timing
+- an observer should be unaware of registry specific types
+- an observer should be chain agnostic
+- an observer should interact with a registry through an interface
+- an observer can define its own dependency interfaces
+- an observer should use a `Coordinator` to track in-flight state
```
infiloop2, line 16: this seems like a new component worthy of its own page.

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
...
+- all observers used by the same plugin instance should use the same registry
```
infiloop2, line 17: I think we can start with the assumption that there will be one observer for a plugin and generalise it later when we see a need

docs/300_OBSERVER.md
EasterTheBunny: Correct. It can be different types or the plugin can have multiple observers. It's all the same. The `Plugin` is responsible to managing rounds, building reports, and feeding accepted reports/transmitted reports back to the observers.

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Filter
+A `Filter` tracks the pending state of upkeeps from the perspective of a single
+node. Once an upkeep is accepted into a report, the filter should indicate
```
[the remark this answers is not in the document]
EasterTheBunny: I was thinking Chris was doing that internally already. Expanding the filter in that way can introduce the idea of middleware in reading observations, whether at the plugin level or at the observer level. ¯\_(ツ)_/¯

docs/300_OBSERVER.md
```diff
@@ -0,0 +1,76 @@
...
+
+```
+type Observer interface {
+    // Observe provides a list of Eligible upkeep identifiers and the `BlockKey`
+    // they were checked at
+    Observe(context.Context) (BlockKey, []UpkeepIdentifier, error)
+    // MakeReport produces a report from upkeep keys; this is a function on the
+    // Observer because on-chain checks will need to be run as well as cache
+    // hits on the Observer itself.
+    MakeReport(context.Context, []UpkeepKey) ([]byte, int, error)
```
infiloop2, line 38: intuitively not sure this makes sense on an observer. This seems like a separate task of merging observations that observers from different nodes provide

docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
...
+}
+```
+
+### May Implement
```
infiloop2, line 41: not sure i follow the difference between must and may implement. Do you have any example in mind for a registry which will not implement this?

docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
...
+
+The following is an example of a `Registry` interface implementation that a 
+polling `Observer` would require. 
+
+```
+// PollingRegistry is intended to provide functions required by a polling
+// observer.
+type PollingRegistry interface {
```
infiloop2, line 49: I thought observers were different types: polling / trigger etc. What does polling in context to registry mean?

docs/500_ENCODER.md
```diff
@@ -0,0 +1,50 @@
+# What is an Encoder?
+The responsibility of an encoder is to interpret raw `PointState` values and
+provide specific operations that cannot be done purely on the raw values. One
+example would be validation of `PointState` raw values.
```
infiloop2, line 4: nit. cleanup PointState

docs/500_ENCODER.md
```diff
@@ -0,0 +1,50 @@
...
+
+An `Encoder` should ideally not contain any state and should simply be an
+interpreter. All functions should be pure.
+
+An `Encoder` can be used by an `Observer`, but in that case the `Observer`
+and the `Plugin` should share the same `Encoder`. An `Encoder` typically maps to
+a single contract instance or registry.
+
+Ideally an encoder should not maintain any state and should be able to be 
+represented with static functions instead of reference functions.
```
infiloop2, line 14: can we enforce this? I don't know about go programming paradigms, but in other languages encode would be an abstract class instead of an object

infiloop2: maybe this makes more sense as a util library?

docs/300_OBSERVER.md
EasterTheBunny: I agree. Instead I kept the `CheckUpkeep` concept and moved the report building to the encoder. Therefore the separation of responsibility is as follows:

Validate and reduce incoming observations: Plugin
Run check upkeep verification: Observer (provides caching) -> Registry
Build report from results: Encoder

docs/700_CONFIG.md
```diff
@@ -0,0 +1,18 @@
...
+
+```
+type IConfig interface {
+    // Has simply indicates that the config has a specific value or not
+    Has(string) bool
+    // Get provides the means to get any value of any type from a configuration
+    // and an error can indicate type issues
+    Get(string, interface{}) error
+}
+```
```
infiloop2, line 14: Why do we need an abstraction here instead of a fixed type?

docs/800_TRANSMITS.md
```diff
@@ -0,0 +1,24 @@
+# Report Coordinator
+
+There are two methods to be supported for confirming transmits made it to chain.
+
+1. check that any key from a `Report` has a perform log received by the node
+2. (future) check that an OCR transmit log was received
```
infiloop2, line 6: this has separate design considerations that I would recommend separating out from here

docs/350_REGISTRY.md
```diff
@@ -0,0 +1,55 @@
...
+
+A `Registry` should be responsible for contract specific behavior and consolidate
+the outcomes of that specific behavior into results that an `Observer` can 
+consume.
+
+To maintain an abstraction between an `Observer`, `Encoder`, and a `Registry`, 
+the return type of a `Registry` should be a wrapper for an `interface{}` type.
+
+```
+type UpkeepResult interface{}
```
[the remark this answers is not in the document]
EasterTheBunny: This is where it all gets a bit sticky. The contents of an upkeep result can change from one contract version to the next, but the only components that need to know the internals are the registry and the encoder. For everything else, `UpkeepResult` is just passed through.

docs/300_OBSERVER.md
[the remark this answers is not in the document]
infiloop2: I think there are separate concerns. Will discuss more offline

docs/350_REGISTRY.md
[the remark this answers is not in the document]
infiloop2: > The contents of an upkeep result can change from one contract version to the next

Don't think we have any use cases for this yet

docs/300_OBSERVER.md
EasterTheBunny: For now, I think that is what we will do. Looping through an array of observers vs reading a single one won't be much different.

docs/350_REGISTRY.md
EasterTheBunny: `GetActiveUpkeepIDs` is specific to a polling observer because it's the only case where all upkeep ids are needed at once. A log trigger variant might need some log topic to upkeep id mappings instead or a cron trigger variant might need a cron config to upkeep id mapping.

docs/350_REGISTRY.md
EasterTheBunny: A polling registry provides functions required for constant polling like the head ticker and getting all ids. Not all observer types will need these.

docs/500_ENCODER.md
EasterTheBunny: Yes, but it wouldn't be very goish. In most cases a function would just accept an interface and the implementation would decide if it was a static or pointer reference. A check **could** be done to ensure the struct passed in matching the interface was not a pointer. I would not advise this as it really seems like a strong anti-pattern.

docs/500_ENCODER.md
```diff
@@ -0,0 +1,50 @@
...
+    // to be used by an `Encoder`.
+    EncodeReport([]UpkeepResult, ...Config) ([]byte, error)
+    // EncodeUpkeepIdentifier should take the output of a registry check result
+    // as input and return an UpkeepIdentifier
+    EncodeUpkeepIdentifier(UpkeepResult) (UpkeepIdentifier, error)
+    // KeysFromReport extracts all upkeep keys from a report byte array
+    KeysFromReport([]byte) ([]UpkeepKey, error)
+    // Eligible takes an upkeep result to provide an abstraction per registry;
+    // returns whether an upkeep is eligible or not along with any decoding
+    // or type conversion errors encountered
```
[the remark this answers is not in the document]
EasterTheBunny: The idea of an encoder is really just a large grouping of functions that satisfy the need of translating general types to specific types and vice-versa. In the re-architecture version of this, the interface was much smaller.

docs/700_CONFIG.md
EasterTheBunny: I'll think more about this one. I wasn't completely sold on it myself. The problem is that configurations change a lot over time and ideally we need to version the configs just as much as we need to version the observations. Each node could get a config that it isn't expecting. If we go with types like `ConfigV1` and `ConfigV2`, functions that need a config would need to accept `interface{}` instead which doesn't provide type safety.

This design provides necessary functions while any specific config manages its own parsing and validation. It's more akin to getting query parameters in an API handler.

docs/350_REGISTRY.md
infiloop2: interesting, this needs more thought. Even with log/cron triggers we'll probably need to do balance / pause / cancelled checks so might need this. 

I guess we can explore this later when we design those specific triggers, do you think for now it would affect the refactor design much?

docs/500_ENCODER.md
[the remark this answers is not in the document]
infiloop2: Not sure i follow completely. Why not have these functions on the type itself `ValidateUpkeepKey`, `ValidateBlockKey` etc?

docs/100_OVERVIEW.md
infiloop2: I think it's fine if our implementation achieves this, my question was more about if we need to have this as explicit goal

docs/350_REGISTRY.md
infiloop2: Do yo envision different types of registry which are not polling?

docs/700_CONFIG.md
infiloop2: We need to think about config management and how that will evolve in future. A simple way to start would be that every new config should have a default value so we don't need different types

PR #95, opened 2023-03-16
Add a new encoder package

ferglor:
https://smartcontract-it.atlassian.net/browse/AUTO-2397

In this PR, we're introducing a new Encoder interface, and an implementation of that interface

This implementation is essentially an amalgamation of existing functionality scattered throughout the codebase

## Questions

- The definition of `EncodeReport` accepts a list of `Config` objects, though when implementing this function, I didn't use any `Config` instances - should we remove the `Config` list parameter, or is it expected that other implementations will use these?
- The concrete implementation of the first encoder is under a `v1` package - my understanding is that we'll have one encoder per contract, but I wasn't sure what the existing contract is called, or what would be a more suitable package name, so I defaulted on to `v1` for now, but will change this as soon as we identify a better name
- Some of the implementations are really simple; I'm not sure if some of the functions should be more involved, or if they need extra validation etc


encoder/v1/encoder.go
```diff
@@ -0,0 +1,218 @@
...
+
+	var baseValuesIdx int
+	for i, rpt := range toReport {
+		if rpt.CheckBlockNumber > uint32(baseValuesIdx) {
+			baseValuesIdx = i
+		}
+	}
+
+	fastGas := toReport[baseValuesIdx].FastGasWei
+	if fastGas == nil {
```
ferglor, line 133: I added the nil checks for FastGasWei and LinkNative because the packer panics if either of these values are nil (discovered in testing)

encoder/v1/encoder.go
```diff
@@ -0,0 +1,218 @@
...
+	UnpackIntoMap(v map[string]interface{}, data []byte) error
+}
+
+type encoder struct {
+	reportKeys []string
+	packer     reportPacker
+}
+
+func NewEncoder() *encoder {
+	arguments := abi.Arguments{
```
ferglor, line 39: Right now the constructor defines the ABI arguments, but we could allow the caller to specify and pass in the arguments instead? But I'm not sure how different versions of the encoder might behave so I've kept it like this for now

encoder/v1/encoder.go
```diff
@@ -0,0 +1,218 @@
...
+	errUpkeepKeyNotParsable = fmt.Errorf("upkeep key not parsable")
+)
+
+type wrappedPerform struct {
+	CheckBlockNumber uint32   `abi:"checkBlockNumber" json:"checkBlockNumber"`
+	CheckBlockhash   [32]byte `abi:"checkBlockhash" json:"checkBlockhash"`
+	PerformData      []byte   `abi:"performData" json:"performData"`
+}
+
+type reportPacker interface {
```
ferglor, line 28: Defining this reportPacker interface makes testing easier, since we can mock the behaviour of the pack and unpack operations

encoder/v1/encoder.go
```diff
@@ -0,0 +1,218 @@
...
+// EncodeReport encodes a report from a list of upkeep results
+func (e *encoder) EncodeReport(toReport []types.UpkeepResult, _ ...encoders.Config) ([]byte, error) {
+	if len(toReport) == 0 {
+		return nil, nil
+	}
+
+	var baseValuesIdx int
+	for i, rpt := range toReport {
+		if rpt.CheckBlockNumber > uint32(baseValuesIdx) {
+			baseValuesIdx = i
```
ferglor, line 128: I find this logic surprising; is `CheckBlockNumber` the block number on which the check was performed? If so, won't the largest `CheckBlockNumber` we encounter always be far greater than the number of `UpkeepResult`s that we pass? I think the main reason I found this surprising was because we're comparing block numbers with slice indices?

encoder/encoder.go
```diff
@@ -0,0 +1,19 @@
+package encoder
+
+import (
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+)
+
+type Config struct{}
```
ferglor, line 7: This struct isn't actually being used in the encoder implementation yet, it's just here to satisfy the interface

encoder/encoder.go
```diff
@@ -0,0 +1,19 @@
...
+
+type Encoder interface {
```
EasterTheBunny, line 9: One pattern that I want to avoid is defining large interfaces in general and defining interfaces in the package in which they are implemented. I prefer defining interfaces in the packages in which they are consumed and the smaller the better.

If this is temporary just for tracking progress, disregard my comments. :)

encoder/v1/encoder.go
EasterTheBunny: A pattern I was toying with was using composure a bit more.

```
package encode

type BigIntBlockKey struct {}

func (k BigIntBlockKey) ValidateBlockKey(BlockKey) (bool, error) {
    // ... apply validation logic
}
```

Which allows for something like this to be done elsewhere:

```
package somethingelse

type V1ContractEncoder struct {
    encode.BigIntBlockKey
    // ... other composed structs
}

func (e *V1ContractEncoder) EncodeReport(UpkeepResult, Config) ([]byte, error) {
    // ... implementation for custom contract
    // it is far more likely that each contract will have a different report encoding
    // than each contract having different BlockKey validation
}
```

encoder/v1/encoder.go
EasterTheBunny: Nice!

encoder/v1/encoder.go
EasterTheBunny: I'll have to look at this more. I wasn't aware that there was a relationship between the `number` of results and the `CheckBlockNumber` of each.

EasterTheBunny (commented):
Added a few thoughts to help break things apart a little. The design so far gives the idea that an `Encoder` is one thing, but I think it might be able to be composed of many little things. That way each version implementation might get simpler with less code duplication.

encoder/encoder.go
ferglor: Yeah totally, I'm happy to move these interfaces into their consuming packages when we get to integrating these new components, but I actually like having to import from the interface package since I find it forces you to consider the relationship between packages more? But totally, these will be moved out to wherever makes sense later on 👍

encoder/v1/encoder.go
infiloop2: can we please fix this logic. It does not make sense to compare checkBlockNumbers with index values

PR #96, opened 2023-03-21
Chore/refactor polling observer

EasterTheBunny:
Used the original polling service to create an Observer that matched the
necessary polling functionality. Some areas were updated such as handling
required services, start, stop, panics, and more.

### DIFF
--- a/internal/util/recoverable.go
+++ b/internal/util/recoverable.go
@@ -1,0 +1,104 @@
+package util
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"runtime/debug"
+	"sync"
+	"time"
+)
+
+var (
+	ErrServiceStopped = fmt.Errorf("service stopped")
+)
+
+type Doable interface {
+	Do() error
+	Stop()
+}
+
+func NewRecoverableService(svc Doable, logger *log.Logger) *RecoverableService {
+	ctx, cancel := context.WithCancel(context.Background())
+	return &RecoverableService{
+		service: svc,
+		stopped: make(chan error, 1),
+		log:     logger,
+		ctx:     ctx,
+		cancel:  cancel,
+	}
+}
+
+type RecoverableService struct {
+	mu      sync.Mutex
+	running bool
+	service Doable
+	stopped chan error
+	log     *log.Logger
+	ctx     context.Context
+	cancel  context.CancelFunc
+}
+
+func (m *RecoverableService) Start() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if m.running {
+		return
+	}
+
+	go m.serviceStart()
+	m.running = true
+}
+
+func (m *RecoverableService) Stop() {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if !m.running {
+		return
+	}
+
+	m.service.Stop()
+	m.cancel()
+	m.running = false
+}
+
+func (m *RecoverableService) serviceStart() {
+	for {
+		select {
+		case err := <-m.stopped:
+			// restart the service
+			if err != nil && errors.Is(err, ErrServiceStopped) {
+				<-time.After(10 * time.Second)
+				m.run()
+			}
+		case <-m.ctx.Done():
+			return
+		}
+	}
+}
+
+func (m *RecoverableService) run() {
+	go func(s Doable, l *log.Logger, chStop chan error, ctx context.Context) {
+		defer func() {
+			if err := recover(); err != nil {
+				if l != nil {
+					l.Println(err)
+					l.Println(string(debug.Stack()))
+				}
+
+				chStop <- ErrServiceStopped
+			}
+		}()
+
+		err := s.Do()
+
+		if l != nil {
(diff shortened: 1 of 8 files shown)


internal/util/recoverable.go
```diff
@@ -0,0 +1,113 @@
...
+	mu      sync.Mutex
+	running bool
+	service Doable
+	stopped chan stoppedService
+	log     *log.Logger
+	ctx     context.Context
+	cancel  context.CancelFunc
+}
+
+func (m *RecoverableService) Start(_ context.Context) error {
```
ferglor, line 43: It looks like we never return an error from this function, only nil, so should we just change the signature to `func (m *RecoverableService) Start(_ context.Context)`? But it also looks like the context isn't used so we could drop it too?

internal/util/recoverable.go
```diff
@@ -0,0 +1,113 @@
...
+	m.running = false
+}
+
+func (m *RecoverableService) serviceStart() {
+	for {
+		select {
+		case st := <-m.stopped:
+			// restart the service
+			if st.Err != nil && errors.Is(st.Err, ErrServiceStopped) {
+				<-time.After(10 * time.Second)
```
ferglor, line 75: Should the rest period between restarts be configurable?

internal/util/recoverable.go
```diff
@@ -0,0 +1,113 @@
...
+		ctx:     ctx,
+		cancel:  cancel,
+	}
+}
+
+type RecoverableService struct {
+	mu      sync.Mutex
+	running bool
+	service Doable
+	stopped chan stoppedService
```
ferglor, line 37: Why is stopped a `chan stoppedService` and not just a `chan error`? I might be missing something but it looks like when we read a `stoppedService` off the channel, we only want to access the error it wraps?

internal/util/recoverable.go
```diff
@@ -0,0 +1,113 @@
...
+		case <-m.ctx.Done():
+			return
+		}
+	}
+}
+
+func (m *RecoverableService) run() {
+	go func(s Doable, l *log.Logger, chStop chan stoppedService, ctx context.Context) {
+		defer func() {
+			if err := recover(); err != nil {
```
ferglor, line 87: Just nitpicking here, but could we rename this `err` to `r` or `rec`, since it's not really an error?

internal/util/result.go
```diff
@@ -0,0 +1,19 @@
...
+	Failures  int
+	Err       error
+}
+
+func (r *Results) Total() int {
+	return r.Successes + r.Failures
+}
+
+func (r *Results) SuccessRate() float64 {
+	return float64(r.Successes) / float64(r.Total())
```
ferglor, line 14: Should we protect against division by zero in both the `SuccessRate` and `FailureRate` functions, with something like:

```
if r.Total() == 0 {
  return 0
}
```

In both implementations?

pkg/polling.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+	}
+}
+
+// processLatestHead performs checking upkeep logic for all eligible keys of the given head
+func (bso *PollingObserver) processLatestHead(ctx context.Context, blockKey types.BlockKey) {
+	var keys []types.UpkeepKey
+	var err error
+
+	{
```
ferglor, line 225: Why are we scoping this?

internal/util/recoverable.go
EasterTheBunny: Maybe. It might also work best as an exponential backoff. I really would like to ensure that hard errors are exposed from this type of service failure. Maybe even go as far as either panicing or throwing errors from the observer. That would be a good bit to figure out.

internal/util/recoverable.go
EasterTheBunny: To be honest, I don't know why I did that. You're right though. No need for the wrapper.

internal/util/recoverable.go
EasterTheBunny: According to the docs, it is an error. The return from `recover()` is the error passed to `panic()`. At least that's my understanding...

```
The recover built-in function allows a program to manage behavior of a panicking goroutine. Executing a call to recover inside a deferred function (but not any function called by it) stops the panicking sequence by restoring normal execution and retrieves the error value passed to the call of panic. If recover is called outside the deferred function it will not stop a panicking sequence. In this case, or when the goroutine is not panicking, or if the argument supplied to panic was nil, recover returns nil. Thus the return value from recover reports whether the goroutine is panicking.
```

internal/util/result.go
EasterTheBunny: Makes sense. I'll add that.

pkg/polling.go
EasterTheBunny: No reason. Got into a habit, lol.

internal/util/recoverable.go
EasterTheBunny: It's weird that it returns an `any` type even though its an error. Maybe an `any` type can be passed to `panic()`??? Never looked into that. ¯\_(ツ)_/¯

PR #98, opened 2023-03-29
Add a coordinator interface

ferglor:


internal/keepers/types.go
```diff
@@ -24,7 +24,7 @@ type upkeepService interface {
 	CheckUpkeep(context.Context, ...types.UpkeepKey) (types.UpkeepResults, error)
 }
 
-type filterer interface {
+type Coordinator interface {
```
EasterTheBunny, line 27: The original idea I had for a refactored `Coordinator` was:

```
type Coordinator interface {
    IsPending(UpkeepKey) (bool, error)
    // Accept adds all upkeep keys to an internal filter
    Accept([]UpkeepKey) error
    // IsTransmissionConfirmed is a temporary function that checks if a log has
    // been received for a particular key
    IsTransmissionConfirmed(UpkeepKey) bool
}
```

The function `IsPending` should act much like the former `Filter` method. I'm not sure this needs an error in the return, but I put it there just to be extra complete. An `Observer` only interacts with the `IsPending` method and nothing else.

ferglor: Doesn't the `Filter` function currently compare that the block key of the `UpkeepKey` is after the `TransmitBlockNumber`? Is that what we consider as being "pending" or is there a separate property we should check?

EasterTheBunny: I think the logic is reversed from within the Filter function itself and the new IsPending function. I can't remember. Part of the reason for the name change, to make what it does a bit more clear.

PR #100, opened 2023-04-04
Implement a polling observer (rebased from #96)

ferglor:
I've created this PR as an easier way of rebasing https://github.com/smartcontractkit/ocr2keepers/pull/96

Chainlink PR: https://github.com/smartcontractkit/chainlink/pull/9187


encoder/v1/eligibility_provider.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+)
+
+type eligibilityProvider struct{}
+
+// Eligible returns whether an upkeep result is eligible
+func (p eligibilityProvider) Eligible(result types.UpkeepResult) bool {
+	return result.State == types.Eligible
+}
```
infiloop2, line 12: From the name of the component, not sure why encode cares about eligibility?

encoder/eligibility_provider.go
```diff
@@ -0,0 +1,9 @@
+package encoder
+
+import (
+	"github.com/smartcontractkit/ocr2keepers/pkg/types"
+)
+
+type EligibilityProvider interface {
+	Eligible(types.UpkeepResult) bool
+}
```
infiloop2, line 9: From the name of the component, not sure why encode cares about eligibility?

encoder/v1/encoder.go
```diff
@@ -0,0 +1,154 @@
...
+func (e *encoder) EncodeReport(toReport []types.UpkeepResult, _ ...encoders.Config) ([]byte, error) {
+	if len(toReport) == 0 {
+		return nil, nil
+	}
+
+	var baseValuesIdx int
+	for i, rpt := range toReport {
+		if rpt.CheckBlockNumber > uint32(baseValuesIdx) {
+			baseValuesIdx = i
+		}
```
infiloop2, line 70: I don't fully follow this logic, why do we compare checkBlockNumber with an index value?

infiloop2: It seems we want to use one value for fastgas and link native. I would say we should use the one with highest checkBlockNumber?

ferglor: This was raised in the original PR for this change: https://github.com/smartcontractkit/ocr2keepers/pull/95#discussion_r1138884835 But it still needs to be addressed

internal/keepers/ocr.go
```diff
@@ -53,50 +55,73 @@ func (k *keepers) Query(_ context.Context, _ types.ReportTimestamp) (types.Query
...
-		BlockKey: chain.BlockKey(blockKey.String()),
-	}
+	for _, observer := range k.observers {
+		block, ids, err := observer.Observe()
+		if err != nil {
+			return nil, fmt.Errorf("%w: failed to sample upkeeps for observation: %s", err, lCtx)
+		}
+
+		if block == nil {
+			k.logger.Printf("observed %d upkeep IDs with nil block", len(allIDs))
```
infiloop2, line 75: nit. should this be an error condition?

internal/keepers/ocr.go
```diff
@@ -53,50 +55,73 @@ func (k *keepers) Query(_ context.Context, _ types.ReportTimestamp) (types.Query
...
+	allIDs = shuffleObservations(allIDs, keyRandSource)
+
 	// Check limit
-	if len(identifiers) > observationUpkeepsLimit {
-		identifiers = identifiers[:observationUpkeepsLimit]
+	if len(allIDs) > observationUpkeepsLimit {
+		allIDs = allIDs[:observationUpkeepsLimit]
+	}
+
+	medianBlock := calculateMedianBlock(blocks, k.reportBlockLag)
```
infiloop2, line 100: not sure why we need this logic here? Is this the local aggregation of different observations here? max might make more sense then?

pkg/coordinator/coordinator.go
```diff
@@ -66,28 +73,25 @@ func newReportCoordinator(r types.Registry, s time.Duration, cacheClean time.Dur
 	return c
 }
 
-// Filter returns a filter function that removes upkeep keys that apply to this
-// filter. Returns false if a key should be filtered out.
-func (rc *reportCoordinator) Filter() func(types.UpkeepKey) bool {
-	return func(key types.UpkeepKey) bool {
-		blockKey, id, err := key.BlockKeyAndUpkeepID()
+// IsPending returns false if a key should be filtered out.
```
infiloop2, line 76: umm, should the name/behaviour be opposite?
If a key is pending then it should be filtered out?

internal/keepers/ocr.go
infiloop2: specifically it does not make sense to subtract reportBlockLag here and changes the intended behaviour of this config

infiloop2:
Can you link this to an associated CL repo PR which runs the smoke tests and is needed to get an image for staging

internal/keepers/ocr.go
ferglor: @EasterTheBunny wdyt? Should we return an error at this point? I think I explicitly added this nil check purely for logging purposes, so I can just drop it completely and either always assume we won't get a nil block, OR only append IDs if the block is non nil, OR just return an error on a nil block?

internal/keepers/ocr.go
ferglor: We don't _really_ need a median block here, it's just that each observer returns a block, and so we have to go from having multiple blocks to having a single block, I can change this to use the zeroth block that gets returned? And that way we eliminate the median calculation and the use of reportBlockLag

pkg/observer/polling/observer.go
```diff
@@ -0,0 +1,429 @@
...
+
+	for _, id := range ids {
+		key := o.upkeepProvider.MakeUpkeepKey(bl, id)
+
+		if !o.filterer.IsPending(key) {
+			continue
+		}
+
+		filteredIDs = append(filteredIDs, id)
+	}
```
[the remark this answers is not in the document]
ferglor: Uhhh maybe actually? I used to think filter functionality might have been observer-specific, but I guess right now they only operate on keys so yeah 🤷

PR #101, opened 2023-04-05
Integrate the filter into the polling observer

ferglor:
In this PR, we're integrating the coordinator's filter functionality into the new polling observer implementation


pkg/polling.go
```diff
@@ -112,6 +115,7 @@ type PollingObserver struct {
 	keys     KeyProvider               // provides keys to this block observer
 	coord    KeyStatusCoordinator      // key status coordinator tracks in-flight status
 	shuffler Shuffler[types.UpkeepKey] // provides shuffling logic for upkeep keys
+	filterer keepers.Coordinator       // provides filtering logic for upkeep keys
```
EasterTheBunny, line 118: We might even be able to simplify this interface since an `Observer` only interacts with `IsPending` anyway. While a `Coordinator` is a larger struct that links some things together, we don't need to define the whole component as a dependency in an `Observer`.

pkg/polling.go
```diff
@@ -129,6 +133,11 @@ func (bso *PollingObserver) Observe() (types.BlockKey, []types.UpkeepIdentifier,
 			continue
 		}
 
+		if !bso.filterer.Filter()(key) {
```
EasterTheBunny, line 136: One idea might be to set this up like middleware. For now the only reason we would filter out an upkeep is because it is in a `pending` state. In the future we could want to filter out upkeeps for other reasons such as the api cooldown logic that Chris added to offChain lookup. I'm sure there are more that will come up so it's an idea.

PR #103, opened 2023-04-11
Integrate encoder components into the polling observer

ferglor:


encoder/v1/eligibility_provider.go
```diff
@@ -7,6 +7,6 @@ import (
 type eligibilityProvider struct{}
 
 // Eligible returns whether an upkeep result is eligible
-func (p eligibilityProvider) Eligible(result types.UpkeepResult) (bool, error) {
-	return result.State == types.Eligible, nil
+func (p eligibilityProvider) Eligible(result types.UpkeepResult) bool {
```
EasterTheBunny, line 10: I want to convert `UpkeepResult` to type `interface{}` eventually. To maintain simplicity in the refactor, we can keep it as is for now and come back to it later.

PR #104, opened 2023-04-13
Integrate polling observer

ferglor:
This PR is aimed at integrating the new polling observer into the keepers plugin

Right now, the observation phase has been updated to call the new observers one by one, and consolidate the results into upkeep results

A solution for how we should integrate these pollers into the report phase is a WIP

### DIFF
--- a/pkg/coordinator/coordinator.go
+++ b/pkg/coordinator/coordinator.go
@@ -30,7 +30,7 @@ import (
 )
 
 type Coordinator interface {
-	Filter() func(types.UpkeepKey) bool
+	IsPending(types.UpkeepKey) bool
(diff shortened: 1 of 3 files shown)


pkg/coordinator/coordinator.go
```diff
@@ -28,6 +29,12 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/pkg/util"
 )
 
+type Coordinator interface {
+	Filter() func(types.UpkeepKey) bool
```
EasterTheBunny, line 33: We should resolve what we're doing with this function. I think we can probably merge all the PRs together to move forward a bit, but filtering keys should be the next focus.

PR #106, opened 2023-04-26
pass mercurylookup boolean to checkUpkeep

FelixFan1992:


RyanRHall:
@FelixFan1992 what is the purpose of this boolean? It seems like we're expecting the caller to tell the registry wrapper if it should handle a mercury lookup or not, but my understanding was that only mercury server will know if an upkeep is allowed access or not.

pkg/types/types_test.go
```diff
@@ -75,6 +78,7 @@ func TestDecode(t *testing.T) {
 			GasOverheadPerUpkeep: 300_000,
 			MaxUpkeepBatchSize:   1,
 			ReportBlockLag:       0,
+			MercuryLookup:        false,
```
infiloop2, line 81: not sure if already covered, but could you ensure that the missing value case is tested? (false being the default)

FelixFan1992: i believe it's already covered in this test where all fields are missing and the expected default is false.
another similar test is the one above where only MercuryLookup is missing and expected value is false.

FelixFan1992:
Tested with chaincli, set config with MercuryLookup true and log the bool value

EasterTheBunny (approved):

PR #107, opened 2023-05-03
Log trigger observer

amirylm:
**NOTE** WIP

## Motivation

The introduction of new trigger type (log trigger) require some changes in our current design so it would fit side by side with other, existing (or future) trigger types.

`Observer` is a trigger agnostic interface for upkeeps, it orchestrates the processing of upkeeps and the exposure of triggered upkeeps for the OCR plugin.

## Description

This PR introduce the new `Observer` interface that was agreed in design phase, and the corresponding implementation of `LogTriggerObserver` that is responsible for log trigger upkeeps.

**NOTE** polling observer for conditional upkeeps will be aligned separately in the future

The following components were added:
- `LogTriggerObserver`
- `Executer` that is used for upkeeps execution by all observers
- Queue/pool to serve the results async to the OCR plugin.

### DIFF
--- a/pkg/observer/observer.go
+++ b/pkg/observer/observer.go
@@ -29,9 +29,9 @@ type ObserverV2[Tick any] interface {
 	// Verify(ctx context.Context, keys ...UpkeepKey) (UpkeepResults, error)
 }
 
-// RegsiterObserver manages a pair of observer/ticker, upon each tick the observer is called
+// RegisterObservers manages a pair of observer/ticker, upon each tick the observer is called
 // with the data received from the channel.
-func RegsiterObservers[T any](
+func RegisterObservers[T any](
 	pctx context.Context,
 	cn <-chan T,
 	observers ...ObserverV2[T],
--- a/pkg/observer/observer.go
+++ b/pkg/observer/observer.go
@@ -15,3 +15,69 @@ type Observer interface {
 	SetSamplingRatio(ratio ratio.SampleRatio)
 	SetMercuryLookup(mercuryLookup bool)
 }
+
+// Observer orchestrates the processing of upkeeps and the exposure of triggered upkeeps for the OCR plugin.
+// Tick is a generic input the service gets from the corresponding ticker (block/time ticker)
+// that invokes the observer.
+type ObserverV2[Tick any] interface {
+	// Process execute upkeeps, triggered by a generic ticker
+	Process(ctx context.Context, t Tick)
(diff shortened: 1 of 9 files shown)


pkg/types/types.go
```diff
@@ -44,6 +44,15 @@ type Registry interface {
 	CheckUpkeep(context.Context, bool, ...UpkeepKey) (UpkeepResults, error)
 }
 
+// Executer provides a runtime layer for parallelized execution of upkeeps,
+// with a cache for executed/pending upkeeps.
+// It acts as a mediator between observers and registry contract / mercury lookup.
+type Executer interface {
+	// Run executes the pipeline for the given upkeeps and collects the results.
+	// checkData is changed based on trigger.
+	Run(ctx context.Context, upkeepKeys []UpkeepKey, checkData [][]byte) (UpkeepResults, error)
```
EasterTheBunny, line 53: For the return, let's do []types.UpkeepResult instead of UpkeepResults unless we really need to extend the base array with some functionality.

Also, how should we handle errors here? If one upkeep check fails due to a 404 on mercury, do we run a retry on the upkeep? How soon do we fail for any specific upkeep?

pkg/observer/observer.go
```diff
@@ -15,3 +15,69 @@ type Observer interface {
...
+	// Propose returns queued upkeep results, called from the OCR plugin upon observation
+	Propose(ctx context.Context) ([]types.UpkeepResult, error)
+	// Verify executes the given upkeeps
+	// decision: not needed for pure reporting
+	// Verify(ctx context.Context, keys ...UpkeepKey) (UpkeepResults, error)
+}
+
+// RegsiterObserver manages a pair of observer/ticker, upon each tick the observer is called
+// with the data received from the channel.
+func RegsiterObservers[T any](
```
EasterTheBunny, line 34: Misspelling: `RegisterObservers`

EasterTheBunny (commented):
Just some quick comments on the WIP. :)

pkg/types/types.go
amirylm: Some retry strategy could work, but we need to distinguish errors that are retry-able.
  In any case, currently the [error handling](https://github.com/smartcontractkit/chainlink/blob/develop/core/services/ocr2/plugins/ocr2keeper/evm/registry.go#L530) and mercury lookup are bound to the `EvmRegistry` so it might be better to handle things is that level.

PR #111, opened 2023-05-12
Log upkeep config

amirylm:


EasterTheBunny (approved):

pkg/types/upkeep_config.go
```diff
@@ -0,0 +1,86 @@
...
+	ErrTopicPrefix           = errors.New("invalid topic: prefixed with 0x")
+)
+
+func (cfg *LogUpkeepConfig) Validate() error {
+	cfg.defaults()
+
+	if len(cfg.Address) == 0 {
+		return ErrContractAddrIsMissing
+	}
+	if n := len(cfg.Topic); n == 0 {
```
ferglor, line 53: 🏄 If `n` isn't being used this can just be `if len(cfg.Topic) == 0`?

pkg/types/upkeep_config.go
```diff
@@ -0,0 +1,86 @@
...
+		return ErrTopicIsMissing
+	}
+	if strings.Index(cfg.Topic, "0x") == 0 {
```
ferglor, line 56: 🏄 `if strings.HasPrefix(cfg.Topic, "0x")` might be clearer?

pkg/types/upkeep_config.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// TODO: implement ConditionalUpkeepConfig
+
+// LogUpkeepConfig holds the settings for a log upkeep
+type LogUpkeepConfig struct {
+	// Address is required, contract address w/ 0x prefix
+	Address string `cbor:"a"`
+	// Topic is required, 32 bytes, w/o 0x prefixed
+	Topic string `cbor:"sig"`
```
ferglor, line 30: 🏄 Why `sig` and not `t`? Just curious 🧐

pkg/types/upkeep_config.go
```diff
@@ -0,0 +1,86 @@
...
+func (cfg *LogUpkeepConfig) Encode() ([]byte, error) {
+	return cbor.Marshal(cfg)
+}
+
+func (cfg *LogUpkeepConfig) Decode(raw []byte) error {
+	return cbor.Unmarshal(raw, cfg)
+}
+
+func (cfg *LogUpkeepConfig) defaults() {
+	if len(cfg.Address) > 0 && strings.Index(cfg.Address, "0x") != 0 {
```
ferglor, line 72: 🏄 HasPrefix could go here too?

pkg/types/upkeep_config_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}{
+		{
+			name: "empty",
+			in:   "",
+			want: "00000000000000000000000000000000",
+		},
+		{
+			name: "short",
+			in:   "129",
+			want: "00000000000000000000000000000129",
```
ferglor, line 78: 🏄 Would it be worth capturing what the output is for a long input string, greater than 32 characters? (more so for documentation purposes rather than actually expecting this to happen)

pkg/types/upkeep_config.go
```diff
@@ -0,0 +1,86 @@
...
+		cfg.Address = fmt.Sprintf("0x%s", cfg.Address)
+	}
+	if len(cfg.Filter1) > 0 && len(cfg.Filter1) < 32 {
+		cfg.Filter1 = zeroPadding(cfg.Filter1)
+	}
+	if len(cfg.Filter2) > 0 && len(cfg.Filter2) < 32 {
+		cfg.Filter2 = zeroPadding(cfg.Filter2)
+	}
```
ferglor, line 80: And Filter3 is not needed here?

ferglor (approved):

pkg/types/upkeep_config.go
amirylm: length check was extracted to its own function

PR #114, opened 2023-05-23
Spike - self cleaning caches

ferglor:
In this spike, we're moving the cache cleaner into the cache itself. This means that, wherever we use a cache, we no longer need to instantiate and run a cache cleaner alongside the cache. With this code, a cache is responsible for running its own cache cleaner internally. The parent component that uses the cache will still be responsible for stopping the cache cleaner, which can be done by calling `cache.Stop()`.

### How can we make this responsibility more explicit?

Right now, when we create a cache, it might not be clear to engineers new to the code base, that we need to explicitly stop the cache. It's also easy to forget to do this, since you can just create a cache, set it as a property on a parent component, and neglect to add `cache.Stop()` to the parent component's `Stop()` implementation.

One approach I was think of, was to return a `Stop()` function from the constructor of the cache, so we'd have something like this:

```golang
cache, stopFn := util.NewCache[types.UpkeepResult](cacheExpire, cacheClean)
```

We'd then add the stopFn to a slice of stop functions, or set it as an extra field on the parent component. This would nudge developers into realizing that they have to do something with this stop function, which might provide a greater level of assurance that caches (and even other managed components like the cache) would be stopped as required.

Corresponding core repo PR: https://github.com/smartcontractkit/chainlink/pull/9321


pkg/util/cache.go
```diff
@@ -26,6 +19,21 @@ type Cache[T any] struct {
 	defaultExpiration time.Duration
 	mu                sync.RWMutex
 	data              map[string]CacheItem[T]
+	cleaner           *IntervalCacheCleaner[T]
+}
+
+func NewCache[T any](expiration, clean time.Duration) *Cache[T] {
```
EasterTheBunny, line 25: What if, to keep the underlying components separately testable, this self-cleaning cache was it's own struct? It would just be composed of both parts, the cache and the cleaner, and would provide the start and stop methods you mention.

ferglor: Yeah we can do! The reason I integrated it this way was because, we seem to always accompany a cache with a cache cleaner in the code, and so it made sense to combine them into a single entity. I guess a cache without a cleaner would need to come with a warning that it would be prone to memory leaks?

PR #116, opened 2023-05-26
Chore/update docs

EasterTheBunny:


docs/diagrams/log_trigger_ticker.md
```diff
@@ -0,0 +1,40 @@
...
+sequenceDiagram
+    participant TK as Ticker[T any]
+    participant OB as Observer[T any]
+    participant R as Registry
+    %% TODO: participant Coordinator
+    participant RN as Runner
+    participant Q as Queue
+    
+    TK->>OB: Process(ctx, T)
+    OB->>R: GetActiveUpkeeps(T)
```
amirylm, line 22: just use `getActiveUpkeepIDsByType`, we can reduce the note

docs/diagrams/sampling_ticker.md
```diff
@@ -0,0 +1,43 @@
...
+sequenceDiagram
+    participant TK as Ticker[T any]
+    participant OB as Observer[T any]
+    participant R as Registry
+    participant RN as Runner
+    participant C as Coordinator
+    participant S as SampleStager
+    
+    TK->>OB: Process(ctx, T)
+    OB->>R: GetActiveUpkeeps(T)
```
amirylm, line 17: just use `getActiveUpkeepIDsByType`, we can reduce the note

docs/ELIGIBILITY_FLOW.md
```diff
@@ -0,0 +1,134 @@
...
+```golang
+type UpkeepIdentifier []byte
+
+type ConfiguredUpkeep struct {
+    // ID uniquely identifies the upkeep
+    ID UpkeepIdentifier
+    // Type is the event type required to initiate the upkeep
+    Type int
+    // Config is configuration data specific to the type
+    Config interface{}
```
amirylm, line 35: `Config` refers to to the upkeep config? would be nice to use actual type here (generics) rather than interface

docs/ELIGIBILITY_FLOW.md
```diff
@@ -0,0 +1,134 @@
...
+check data, etc.
+
+This can potentially be a combination of both `ConfiguredUpkeep` and `T`:
+
+```golang
+type UpkeepPayload struct {
+    // Upkeep is all information that identifies the upkeep
+    Upkeep ConfiguredUpkeep
+    // Tick is the event that triggered the upkeep to be checked
+    Tick interface{}
```
amirylm, line 52: Would be nice to use actual type here (generic Tick) rather than interface

docs/ELIGIBILITY_FLOW.md
```diff
@@ -0,0 +1,134 @@
...
+```golang
+type CheckResult struct {
+    // UpkeepPayload is the original payload request
+    UpkeepPayload
+    // Retryable indicates whether the result failed or not and should be retried
+    Retryable bool
+    // Eligible indicates whether the upkeep is performable or not
+    Eligible bool
+    // Data is a general container for raw results that pass through the plugin
+    Data interface{}
```
amirylm, line 71: `[]byte` might be better here, then it could be parsed to some actual type on demand

docs/ELIGIBILITY_FLOW.md
```diff
@@ -0,0 +1,134 @@
...
+check data, etc.
+
+```golang
+type UpkeepIdentifier []byte
+
+type ConfiguredUpkeep struct {
+    // ID uniquely identifies the upkeep
+    ID UpkeepIdentifier
+    // Type is the event type required to initiate the upkeep
+    Type int
```
amirylm, line 33: We'll be able to derive the type from the ID, could be redundant as additional field

EasterTheBunny: The idea in this struct is to pack the type and id together in the registry, but only split it apart once in the go code and pass them around as two values.

docs/ELIGIBILITY_FLOW.md
EasterTheBunny: This comes from the registry though, so apart from creating a registry instance per observer, how do we use generics here?

docs/ELIGIBILITY_FLOW.md
EasterTheBunny: Same here. The payload is being passed to the registry. Not sure how to do generics when there are multiple typed values.

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+For the case of node restarts, it was proposed that local caches would be
+cleared exposing a possibility that upkeeps could be reported by a set of 
+restarted nodes resulting in successful (bad) transmits.
+
+Approach #1: maintain recent results in outcome and use as a filter for new
+outcomes. (not shown)
+
+Approach #2: consider the lack of reporting an upkeep as a valid observation.
+f+1 nodes can indicate that an upkeep needs to be performed while f+1 nodes can
+also indicate that an upkeep does not need to be performed. In this case, both
```
infiloop2, line 131: Note: We are not guaranteed to get all observations from n nodes. nodes might have a legitimate delay in sending observation to leader

infiloop2: Also i'm not sure how we will differentiate between a node being down and a node saying an upkeep is not needed

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+
+    // Outcome -------------------
+    for each attributedObservation
+        increment count for check pipeline result
+        update median block number with observation latest
+        add observation id samples to sampled id list
+
+    for each attributedObservation
+        for each flattened check pipeline result
+            if check pipeline result is not in observation
```
infiloop2, line 170: what if a node is just lagging? Do we delay an upkeep until enough nodes report that upkeep. What is "enough" here

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+                decrement count for check pipeline result
+
+    for each flattened check pipeline result
+        if count is greater than 0
```
infiloop2, line 174: I assume this is a proxy for f+1 quorum? Does this depend on the number of observations a leader gets?

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+    get check pipeline results from ocr staging
+    
+    for each check pipeline result
+        add check pipeline result to observation
+
+    get latest block from sample staging
+    add latest block to observation
+
+    // only eligible upkeeps
+    get sampled upkeep ids from sample staging
```
infiloop2, line 154: does there need to be any sampling here

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+observations would achieve a valid quorum, but produce different outcomes. 
+Comparing the observations against each other, accounting for both the positive 
+*AND* negative, a final unified outcome can be achieved that is resistant to
+node restarts or fresh nodes entering the network. (shown)
+```
+for each round
+    // Observation -------------------
+    for each check pipeline result in previous outcome
+        add check pipeline result to in-flight cache
+        remove check pipeline result from ocr staging
```
infiloop2, line 141: Where do we use the sampling results from outcome

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+            if check pipeline result is not in observation
+                decrement count for check pipeline result
+
+    for each flattened check pipeline result
+        if count is greater than 0
+            add to outcome
+
+    flatten and dedupe sampled ids
+    add sampled ids to outcome
+    add median block to outcome
```
infiloop2, line 179: how are these used?

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+general, it pulls from some collection of staged values, constructs observations,
+and builds reports. Some components hook into this process to feed progress
+data to the eligibility check flows.
+
+For the case of node restarts, it was proposed that local caches would be
+cleared exposing a possibility that upkeeps could be reported by a set of 
+restarted nodes resulting in successful (bad) transmits.
+
+Approach #1: maintain recent results in outcome and use as a filter for new
+outcomes. (not shown)
```
infiloop2, line 127: What about the approach of using in-flight cache which you also mention

docs/PSEUDOCODE.md
```diff
@@ -0,0 +1,191 @@
...
+
+## OCR3 Report Flow
+OCR is allowed to run on a separate cadence from the eligibility check flow. In
+general, it pulls from some collection of staged values, constructs observations,
+and builds reports. Some components hook into this process to feed progress
+data to the eligibility check flows.
+
+For the case of node restarts, it was proposed that local caches would be
+cleared exposing a possibility that upkeeps could be reported by a set of 
+restarted nodes resulting in successful (bad) transmits.
```
infiloop2, line 124: This should be fine for now, the contract can protect against duplicated transmits

PR #121, opened 2023-06-05
Add a result store

ferglor:
In this PR, I'm adding a result store that provides functionality to add, remove, and view all items in the store. Under the hood, the store uses a map of strings to generic type.

### DIFF
--- a/pkg/stores.go
+++ b/pkg/stores.go
@@ -1,0 +1,37 @@
+package ocr2keepers
+
+import "fmt"
+
+type ResultStore[T any] interface {
+	Add(T)
+	Remove(T)
+	View() ([]T, error)
+}
+
+type resultStore[T any] struct {
+	data map[string]T
+}
+
+func NewResultStore[T any]() *resultStore[T] {
+	return &resultStore[T]{
+		data: make(map[string]T),
+	}
+}
+
+func (s *resultStore[T]) Add(result T) {
+	key := fmt.Sprintf("%v", result)
+	s.data[key] = result
+}
+
+func (s *resultStore[T]) Remove(result T) {
+	key := fmt.Sprintf("%v", result)
+	delete(s.data, key)
+}
+
+func (s *resultStore[T]) View() ([]T, error) {
+	var result []T
+	for _, r := range s.data {
+		result = append(result, r)
+	}
+	return result, nil
+}
--- a/pkg/stores.go
+++ b/pkg/stores.go
@@ -18,14 +18,18 @@ func NewResultStore[T any]() *resultStore[T] {
 	}
 }
 
-func (s *resultStore[T]) Add(result T) {
-	key := fmt.Sprintf("%v", result)
-	s.data[key] = result
+func (s *resultStore[T]) Add(results ...T) {
+	for _, result := range results {
+		key := fmt.Sprintf("%v", result)
+		s.data[key] = result
+	}
 }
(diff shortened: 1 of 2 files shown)


pkg/stores_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+type checkResult struct {
+	Retryable bool
+	Data      string
+}
+
+func TestNewResultStore(t *testing.T) {
```
cmalec, line 14: i personally find table driven tests much easier to read and to see what things are being tested. You don't need to change it since we haven't really agreed to this as a team. But maybe if you like the idea we can work on making new tests follow the pattern.
in GoLand you can have it generate the shell for you it looks like this
```
func Test_resultStore_Add(t *testing.T) {
	type args[T any] struct {
		result T
	}
	type testCase[T any] struct {
		name string
		s    resultStore[T]
		args args[T]
	}
	tests := []testCase[ /* TODO: Insert concrete types here */ ]{
		// TODO: Add test cases.
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			tt.s.Add(tt.args.result)
		})
	}
}
```

cmalec (approved):

pkg/stores_test.go
ferglor: Oh yeah me too! I just didn't use them this time because the functionality here is quite limited and linear, like it either adds or it doesn't and it either removes or it doesn't, so I thought a test table would get messy? I usually use test tables for testing functions that return values I guess?

pkg/stores.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func NewResultStore[T any]() *resultStore[T] {
+	return &resultStore[T]{
+		data: make(map[string]T),
+	}
+}
+
+func (s *resultStore[T]) Add(result T) {
+	key := fmt.Sprintf("%v", result)
```
amirylm, line 22: Would be necessary to introduce a key instead of encoding the object that we are saving as a key (or at least use something shorter such as sha256 or md of the data)

pkg/stores.go
```diff
@@ -0,0 +1,37 @@
...
+import "fmt"
+
+type ResultStore[T any] interface {
+	Add(T)
+	Remove(T)
+	View() ([]T, error)
+}
+
+type resultStore[T any] struct {
+	data map[string]T
```
amirylm, line 12: Thread safety?

pkg/stores.go
```diff
@@ -0,0 +1,37 @@
+package ocr2keepers
+
+import "fmt"
+
+type ResultStore[T any] interface {
```
amirylm, line 5: Let's support multiple items per operation for `Add(...T)` and `Remove(...T)`

pkg/stores.go
EasterTheBunny: The scope of this card is just to get a bare-bones result store in place as a required dependency. There is a follow-up card for applying more detail. I'll add this note there for future implementation.

pkg/stores.go
EasterTheBunny: Let's add this to the follow-up card too so it doesn't block forward momentum

EasterTheBunny: I'm good with var args.

PR #122, opened 2023-06-05
Add an eligible post processor

ferglor:
In this PR, we're adding a new post processor that adds eligible results to its underlying store

### DIFF
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -37,4 +37,6 @@ type StaleReportLog struct {
 
 type UpkeepPayload interface{}
 
-type CheckResult interface{}
+type CheckResult interface {
+	IsEligible() bool
+}
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -37,6 +37,6 @@ type StaleReportLog struct {
 
 type UpkeepPayload interface{}
 
-type CheckResult interface {
-	IsEligible() bool
+type CheckResult struct {
+	Eligible bool
 }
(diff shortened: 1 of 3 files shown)


pkg/postprocessors/eligible.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+)
+
+type checkResultAdder interface {
+	Add(ocr2keepers.CheckResult)
+}
+
+type PostProcessor interface {
```
shileiwill, line 17: i guess this interface is not used?

pkg/basetypes.go
```diff
@@ -34,3 +34,7 @@ type StaleReportLog struct {
 	Confirmations   int64
 	TransactionHash string
 }
+
+type CheckResult interface {
```
shileiwill, line 38: interface or struct as in eligible_test.go L13?

shileiwill (approved):

pkg/basetypes.go
EasterTheBunny: You can do a struct here with a property of `Eligible`. The reason for the property instead of a function that returns eligibility is to ease marshaling in the event we need to add these values to an observation or outcome. That was the purpose for defining public structs in the refactor with distinct types. Many of these values will need to be unmarshalled in telemetry.

EasterTheBunny (approved):

shileiwill (approved):
LGTM

PR #124, opened 2023-06-06
AUTO-3261   Generic Observer

cmalec:


pkg/observer.go
```diff
@@ -0,0 +1,83 @@
...
+type UpkeepPayload interface{}
+type CheckResult interface{}
+
+type Tick interface {
+	// GetUpkeeps provides an array of upkeeps scoped to the individual tick
+	GetUpkeeps(ctx context.Context) ([]UpkeepPayload, error)
+}
+
+// Runner2 is the interface for an object that should determine eligibility state
+// (I didn't want to mess with the existing runner and was unsure if we wanted to reuse that)
```
EasterTheBunny, line 14: Good callout. We do want to re-use that and there is a ticket for it as well. I'll add a note in that ticket to unify them both, but first a new function will need to be added to the existing runner and the function signature updated.

pkg/observer.go
```diff
@@ -0,0 +1,83 @@
...
+	PreProcess(context.Context, []UpkeepPayload) ([]UpkeepPayload, error)
+}
+
+// Postprocessor is the general interface for a processing function after checking eligibility status
+type Postprocessor interface {
+	// PostProcess takes a slice of results where eligibility status is known
+	PostProcess(context.Context, []CheckResult) error
+}
+
+type Observer interface {
```
EasterTheBunny, line 33: This isn't needed in this file since the observer interface isn't a dependency of the observer.

EasterTheBunny (changes requested):
small change and I'll update the details of the runner ticket

pkg/observer_test.go
```diff
@@ -0,0 +1,237 @@
...
+type MockPreprocessor struct {
+	mock.Mock
+}
+
+func (m *MockPreprocessor) PreProcess(ctx context.Context, payloads []UpkeepPayload) ([]UpkeepPayload, error) {
+	ret := m.Called(ctx, payloads)
+	return ret.Get(0).([]UpkeepPayload), ret.Error(1)
+}
+
+type MockPostprocessor struct {
```
ferglor, line 39: 🏄 I would have these mock types as unexported but I'm happy with whatever you prefer 👍

ferglor (approved):

pkg/observer_test.go
cmalec: i didnt see a mocks directory. but i can change it later

PR #126, opened 2023-06-06
add retry processor

shileiwill:


pkg/postprocessors/retry_test.go
```diff
@@ -0,0 +1,40 @@
...
+	// Create some check results
+	results := []CheckResult{
+		{Retryable: true},
+		{Retryable: false},
+		{Retryable: true},
+	}
+
+	// Call the PostProcess method
+	err := processor.PostProcess(context.Background(), results)
+	if err != nil {
```
ferglor, line 32: Instead of using checks like `if`, I think it's cleaner in tests to use `assert`, e.g. you can replace this if block with:

assert.Nil(t, err, "PostProcess returned an error: %v", err)

pkg/postprocessors/retry_test.go
```diff
@@ -0,0 +1,40 @@
...
+		t.Errorf("PostProcess returned an error: %v", err)
+	}
+
+	// Verify that the Retry method was called for retryable results
+	if !retryer.retryCalled {
```
ferglor, line 37: Same here:

assert.True(t, retryer.retryCalled, "Retry method was not called for retryable results")

pkg/postprocessors/retry.go
```diff
@@ -0,0 +1,36 @@
+package postprocessors
+
+import (
+	"context"
+)
+
+type CheckResult struct {
```
ferglor, line 7: I think this struct should live in pkg/basetypes.go?

shileiwill:
Rebased on top of most recent develop, run test coverage and result is 100%

EasterTheBunny (approved):
LGTM

pkg/postprocessors/retry.go
```diff
@@ -0,0 +1,30 @@
+package postprocessors
```
EasterTheBunny, line 1: Let's put this package in `pkg/v3/postprocessors`

EasterTheBunny (changes requested):

EasterTheBunny (approved):

PR #127, opened 2023-06-06
Add a basic time ticker

ferglor:
This PR adds a basic time ticker that runs on a timed interval. On each tick, we call `Process` on the supplied observer. We pass a `Tick` instance to the `Process` function. The `Tick` instance references the time of the tick, and a function for retrieving upkeep payloads.

### DIFF
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -40,3 +40,5 @@ type UpkeepPayload interface{}
 type CheckResult struct {
 	Eligible bool
 }
+
+type UpkeepPayload interface{}
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -40,5 +40,3 @@ type UpkeepPayload interface{}
 type CheckResult struct {
 	Eligible bool
 }
-
-type UpkeepPayload interface{}
--- a/pkg/tickers/time.go
+++ b/pkg/tickers/time.go
@@ -28,7 +28,7 @@ type Tick interface {
 	GetUpkeeps(ctx context.Context) ([]ocr2keepers.UpkeepPayload, error)
 }
 
-type getUpkeepsFn func(context.Context, time.Time) ([]ocr2keepers.UpkeepPayload, error)
+type getUpkeepsFn func(context.Context, time.Time) (Tick, error)
 
 type timeTicker struct {
 	interval time.Duration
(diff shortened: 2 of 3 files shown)


pkg/tickers/time.go
```diff
@@ -0,0 +1,74 @@
...
+type getUpkeepsFn func(context.Context, time.Time) (Tick, error)
+
+type timeTicker struct {
+	interval time.Duration
+	ticker   *time.Ticker
+	observer observer
+	getterFn getUpkeepsFn
+}
+
+func NewTimeTicker(interval time.Duration, observer observer, getterFn getUpkeepsFn) *timeTicker {
```
EasterTheBunny, line 40: For now using just the interval is fine, but there will probably be a case where the ticker interval is different from the `Process` timeout. The use case would be time ticks every second but an observe process that is allowed to take 5 seconds to complete. This would result in lost time intervals, but that is expected when `Process` blocks.

pkg/tickers/time.go
```diff
@@ -0,0 +1,74 @@
...
+	go t.Start()
+	return t
+}
+
+func (t *timeTicker) Start() {
+	for {
+		select {
+		case tm := <-t.ticker.C:
+			func() {
+				ctx, cancelFn := context.WithTimeout(context.Background(), t.interval)
```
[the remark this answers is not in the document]
ferglor: Right, but what context should we derive that context from?

pkg/tickers/time.go
ferglor: Okay so should we pass an interval and a timeout duration here, or are we okay to leave it as-is for now?

pkg/basetypes.go
```diff
@@ -38,3 +38,5 @@ type StaleReportLog struct {
 type CheckResult struct {
 	Eligible bool
 }
+
+type UpkeepPayload interface{}
```
shileiwill, line 42: we will need to hold some variables like Hash/ID for this UpkeepPayload, you think it is better to change this interface to struct? The other way around is to have a GetHash() fn, want to hear your thoughts.

ferglor: Yeah we can change it to a struct for sure! I don't need to use any particular properties on it in this PR so I've left it like this, but a future PR can change this as needed 👍

pkg/tickers/time.go
```diff
@@ -0,0 +1,80 @@
...
+	return t
+}
+
+func (t *timeTicker) Start() {
+	for {
+		select {
+		case tm := <-t.ticker.C:
+			func() {
+				ctx, cancelFn := context.WithTimeout(t.ctx, t.interval)
+				defer cancelFn()
```
shileiwill: so this cancelFn is different with the cancelFn in timeTicker as in L50, what are the use cases for both?

shileiwill (approved):
Left a minor comment, looks great.

pkg/tickers/time.go
ferglor: In go, when you create a context, you must specify a parent context from which the new context will be derived. So the cancelFn on line 50 is the parent context, and then on line 61, we create a new child context from the parent context. The child context has a timeout equal to the tick interval, to ensure that the subsequent operations complete before the ticker's next tick. When we call Stop() on this ticker, we cancel the parent context, which in turn cancels the child context.

PR #130, opened 2023-06-07
separated runner types to keep different versions from interacting

EasterTheBunny:


ferglor (approved):
I've built the chainlink core repo against this change without issue 👍

PR #113, opened 2023-05-17
Chore/fully integrate refactor

EasterTheBunny:

PR #131, opened 2023-06-07
Keepers refactor

ferglor:


encoder/v1/encoder.go
```diff
@@ -0,0 +1,155 @@
...
+	if len(toReport) == 0 {
+		return nil, nil
+	}
+
+	var baseValuesIdx int
+	for i, rpt := range toReport {
+		// TODO AUTO-2863
+		if rpt.CheckBlockNumber > uint32(baseValuesIdx) {
+			baseValuesIdx = i
+		}
```
infiloop2, line 71: is it safe to merge with this todo in place?

encoder/v1/validator.go
```diff
@@ -0,0 +1,54 @@
...
+	_, err = v.ValidateUpkeepIdentifier(upkeepIdentifier)
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// ValidateUpkeepIdentifier returns true if the types.UpkeepIdentifier is valid, false otherwise
+func (v validator) ValidateUpkeepIdentifier(identifier types.UpkeepIdentifier) (bool, error) {
```
infiloop2, line 33: should we have a check that it's not greater than uint256.max here too? (not fully sure but i think we had that before?)

encoder/v1/validator.go
```diff
@@ -0,0 +1,54 @@
...
+		return false, fmt.Errorf("upkeep identifier is not a big int")
+	}
+	if identifierInt.Cmp(big.NewInt(0)) == -1 {
+		return false, fmt.Errorf("upkeep identifier is not a positive integer")
+	}
+	return true, nil
+}
+
+// ValidateBlockKey returns true if the types.BlockKey is valid, false otherwise
+func (v validator) ValidateBlockKey(key types.BlockKey) (bool, error) {
```
infiloop2, line 45: should we have a check that it's not greater than uint64.max here too? (not fully sure but i think we had that before?)

encoder/v1/validator.go
infiloop2: we have this in main https://github.com/smartcontractkit/ocr2keepers/blob/main/pkg/chain/upkeep_observation.go#L41-L57 not sure if this check is still present somewhere else

internal/keepers/factory.go
```diff
@@ -1,3 +1,8 @@
+/*
+The keepers package is deprecated and will no longer be expected to function. It
+will be kept for a short period of time for reference. Use functions in the
```
infiloop2, line 3: Not sure the value of keeping this here. It will always be present in git history.

imo would be just cleaner to delete it to prevent confusion / accidental use

encoder/v1/encoder.go
ferglor: Ah so this was fixed in the PR that was referencing the change that was missed out; this component isn't being used anymore, can you confirm @EasterTheBunny

pkg/coordinator/coordinator.go
```diff
@@ -1,109 +1,154 @@
...
 	"github.com/smartcontractkit/ocr2keepers/pkg/util"
 )
 
 var (
-	ErrKeyAlreadyAccepted = fmt.Errorf("key alredy accepted")
-	IndefiniteBlockingKey = chain.BlockKey("18446744073709551616") // Higher than possible block numbers (uint64), used to block keys indefintely
+	DefaultLockoutWindow  = time.Duration(20) * time.Minute
+	DefaultCacheClean     = time.Duration(30) * time.Second
+	ErrKeyAlreadyAccepted = fmt.Errorf("key already accepted")
+	// TODO: still chain specific code
```
infiloop2, line 41: can we have tickets to track all todo's please

pkg/coordinator/coordinator.go
```diff
@@ -1,109 +1,154 @@
...
+func (rc *reportCoordinator) Accept(key ocr2keepers.UpkeepKey) error {
+	blockKey, id, err := rc.encoder.SplitUpkeepKey(key)
 	if err != nil {
 		return err
 	}
 
 	// If a key is already active then don't update filters, but also not throw errors as
 	// there might be other keys in the same report which can get accepted
-	if _, ok := rc.activeKeys.Get(key.String()); !ok {
+	// TODO: key to string again
```
infiloop2, line 144: can we link a ticket please

pkg/coordinator/coordinator.go
```diff
@@ -112,15 +157,27 @@ func (rc *reportCoordinator) Accept(key types.UpkeepKey) error {
...
 }
 
-func (rc *reportCoordinator) IsTransmissionConfirmed(key types.UpkeepKey) bool {
+// IsTransmissionConfirmed returns whether the upkeep was successfully
+// completed or not
+func (rc *reportCoordinator) IsTransmissionConfirmed(key ocr2keepers.UpkeepKey) bool {
 	// key is confirmed if it both exists and has been confirmed by the log
 	// poller
-	confirmed, ok := rc.activeKeys.Get(key.String())
+	confirmed, ok := rc.activeKeys.Get(string(key))
```
infiloop2, line 165: same here, i am not sure I follow why this had to change

pkg/coordinator/coordinator.go
ferglor: This was addressed in the original review here: https://github.com/smartcontractkit/ocr2keepers/pull/113#discussion_r1197163626

pkg/coordinator/coordinator.go
```diff
@@ -325,9 +401,15 @@ func (rc *reportCoordinator) run() {
 		case <-timer.C:
 			startTime := time.Now()
 
-			rc.checkLogs()
+			// limit the check process to 5 times the cadence as an upper bound
+			ctx, cancel := context.WithTimeout(context.Background(), 5*cadence)
+			if err := rc.checkLogs(ctx); err != nil {
```
infiloop2, line 406: Not sure of this behaviour. Previously say there was an overload of logs and it took 10 seconds to fetch logs from DB we would have still done that and cleared the cache slowly. Now we will continuously error out. 

What would be the case here for erroring out after 5s?

encoder/v1/encoder.go
infiloop2: please delete the component if not being used

pkg/coordinator/coordinator.go
infiloop2: so this is intentional to not have string method on a []byte? we shouldn't have the TODO then

fwiw it's not clear from this snippet that it will always be a []byte as it is an interface type. I thought the whole purpose was not to tie the logic with concrete types :upside_down_face:

pkg/encoding/basic.go
```diff
@@ -0,0 +1,141 @@
...
+	components := strings.Split(string(upkeepKey), separator)
+	if len(components) != 2 {
+		return "", nil, fmt.Errorf("%w: missing data in upkeep key", ErrUpkeepKeyNotParsable)
+	}
+
+	return ocr2keepers.BlockKey(components[0]), ocr2keepers.UpkeepIdentifier(components[1]), nil
+}
+
+// ValidateUpkeepKey returns true if the types.UpkeepKey is valid, false otherwise
+func (kb BasicEncoder) ValidateUpkeepKey(key ocr2keepers.UpkeepKey) (bool, error) {
```
infiloop2, line 40: moving the comment here about validation of upper bounds on block key and upkeep id

pkg/observer.go
```diff
@@ -0,0 +1,69 @@
...
+	// Get upkeeps from tick
+	upkeeps, err := tick.GetUpkeeps(ctx)
+	if err != nil {
+		return err
+	}
+
+	var upkeepPayloads []UpkeepPayload
+	// Run pre-processors
+	for _, preprocessor := range o.Preprocessors {
+		upkeepPayloads, err = preprocessor.PreProcess(ctx, upkeeps)
```
infiloop2, line 55: `PreProcess` takes Upkeep Payload as an input while we pass upkeeps here. It is confusing what's the difference between the two entities.

infiloop2: I see `UpkeepPayload` has a type `interface{}` and it seems we are abusing it for upkeepIDs as well as other upkeep Data

pkg/observer.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Postprocessor is the general interface for a processing function after checking eligibility status
+type Postprocessor interface {
+	// PostProcess takes a slice of results where eligibility status is known
+	PostProcess(context.Context, []CheckResult) error
+}
+
+// Runner2 is the interface for an object that should determine eligibility state
+type Runner2 interface {
```
infiloop2, line 24: what is runner1?

pkg/observer.go
infiloop2: Is there potential to use generics here to get better typing?

pkg/observer/polling/observer.go
```diff
@@ -0,0 +1,337 @@
...
+	runner        Runner
+	encoder       Encoder
+	mercuryLookup bool
+}
+
+// Observe implements the Observer interface and provides a slice of identifiers
+// that were observed to be performable along with the block at which they were
+// observed. All ids that are pending are filtered out.
+func (o *PollingObserver) Observe() (ocr2keepers.BlockKey, []ocr2keepers.UpkeepIdentifier, error) {
+	bl, ids := o.stager.get()
```
infiloop2, line 140: is this new component being used in 2.0 or just new for 2.1?

pkg/observer/polling/observer.go
```diff
@@ -0,0 +1,337 @@
...
+	})
+
+	return nil
+}
+
+func (o *PollingObserver) runHeadTasks() error {
+	ch := o.heads.HeadTicker()
+	for {
+		select {
+		case bl := <-ch:
```
infiloop2, line 191: As I understand correctly, it expected that this channel will drop heads while the current one is being processed?

I feel this is a very non trivial assumption and would be good to surface that our more clearly (comments / renaming `HeadTicker` variable to make this clear

pkg/observer/polling/observer.go
```diff
@@ -0,0 +1,337 @@
...
+	// Eligible returns whether or not the upkeep result should be performed
+	Eligible(ocr2keepers.UpkeepResult) (bool, error)
+	// MakeUpkeepKey combines a block and upkeep id into an upkeep key. This
+	// will probably go away with a more structured static upkeep type.
+	MakeUpkeepKey(ocr2keepers.BlockKey, ocr2keepers.UpkeepIdentifier) ocr2keepers.UpkeepKey
+	// SplitUpkeepKey ...
+	SplitUpkeepKey(ocr2keepers.UpkeepKey) (ocr2keepers.BlockKey, ocr2keepers.UpkeepIdentifier, error)
+	// Detail is a temporary value that provides upkeep key and gas to perform.
+	// A better approach might be needed here.
+	Detail(ocr2keepers.UpkeepResult) (ocr2keepers.UpkeepKey, uint32, error)
```
infiloop2, line 38: Not fully understand the need here. As I see it's only being used to fetch the upkeepKey from the result currently?

pkg/observer.go
ferglor: `Runner2` is a type that was introduced as part of the OCR3 fan out work and can be ignored as part of the refactor review; unfortunately the PR for this change was merged into develop before I could get this PR open

pkg/observer/polling/observer.go
infiloop2: if `ocr2keepers.UpkeepResult` was better typed here instead of interface then we wouldn't need this I think

pkg/ocr.go
```diff
@@ -0,0 +1,384 @@
...
+
+	allIDs = shuffleObservations(allIDs, getRandomKeySource(t))
+
+	// limit the total number of observations if over the limit
+	if len(allIDs) > ObservationUpkeepsLimit {
+		allIDs = allIDs[:ObservationUpkeepsLimit]
+	}
+
+	// use the encoder to get the median block from the list of all blocks
+	// median := p.encoder.GetMedian(blocks)
```
infiloop2, line 132: nit. can we clean up unused code if this file is being used in prod

pkg/runner/runner.go
```diff
@@ -0,0 +1,199 @@
...
+				ok, err := o.encoder.Eligible(result)
+				if err != nil {
+					o.logger.Printf("eligibility check error: %s", err)
+					continue
+				}
+
+				if ok {
+					key, _, _ := o.encoder.Detail(result)
+					o.logger.Printf("upkeep ready to perform for key %s", key)
+				}
```
infiloop2, line 174: nit. can we also log the opposite (when an upkeep is not eligible)

We do this in main https://github.com/smartcontractkit/ocr2keepers/blob/main/internal/keepers/service.go#L318-L320

pkg/observer.go
infiloop2: my question was more about why this is called runner2 :sweat_smile: ant not just runner

encoder/v1/encoder.go
ferglor: It's going to be used by the OCR3 changes eventually

pkg/observer.go
ferglor: Haha, okay so there's a Runner component that we use in v2.0. When the OCR3-related PR was merged into develop, the chainlink build was failing because there was an interface change in the Runner component. For this reason, Runner2 was created to sidestep the chainlink build issue. So this is a temporary name for what will eventually be Runner. The problem here is that develop was used as the refactor base branch, but that has now extended into being used as the OCR3 branch.

pkg/coordinator/coordinator.go
```diff
@@ -222,24 +282,28 @@ func (rc *reportCoordinator) checkLogs() {
...
 		}
 	}
+
+	return err
 }
 
 type idBlocker struct {
-	CheckBlockNumber    types.BlockKey
-	TransmitBlockNumber types.BlockKey
+	Encoder             Encoder
```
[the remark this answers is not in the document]
infiloop2: small PR for cleanup https://github.com/smartcontractkit/ocr2keepers/pull/135

pkg/coordinator/coordinator.go
infiloop2: I discussed this with @ferglor and we both think it would be better to have a string interface to avoid implicit assumption of underlying type being bytes. 

@ferglor feel free to put up a PR adding the string interface

pkg/observer/polling/observer.go
EasterTheBunny: This is used in 2.0, not in 2.1

pkg/observer.go
EasterTheBunny: This can be moved to a temporary v2 package just to make the separation clear. We just have to move 2 files

pkg/observer.go
infiloop2: I'm ok with this if explicitly just used in v2

pkg/coordinator/coordinator.go
EasterTheBunny: I like the idea of an upper bound on the time the database can take. Database queries already have a timeout anyway so the only reason for running this function to completion of the query is to just rely on that timeout. We can use that method instead.

pkg/coordinator/coordinator.go
EasterTheBunny: If we do this and pass around interfaces, we break the ability of having static types. We can't have both.

pkg/observer/polling/observer.go
EasterTheBunny: This will be resolved in the next version

infiloop2 (approved):
looks good! thanks for resolving all the comments

anirudhwarrier (approved):

PR #133, opened 2023-06-07
Retry ticker

shileiwill:
This PR is on top of the basic time ticker PR. Need to rebase, retest after the time ticker PR lands.

### DIFF
--- a/pkg/tickers/retry.go
+++ b/pkg/tickers/retry.go
@@ -1,0 +1,94 @@
+package tickers
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+	"github.com/smartcontractkit/ocr2keepers/pkg/util"
+)
+
+const (
+	retryDelay              = 500 * time.Millisecond
+	totalAttempt            = 3
+	attemptsCacheExpiration = 10 * retryDelay // 5 seconds
+)
+
+type retryTicker struct {
+	timeTicker
+	nextRetries     sync.Map // time.Time -> ocr2keepers.UpkeepPayload
+	payloadAttempts *util.Cache[int]
+}
+
+// Retry adds a retryable result to the retryTicker.
+func (rt *retryTicker) Retry(result ocr2keepers.CheckResult) error {
+	payload := result.Payload
+
+	if result.Retryable {
+		attemptCount, ok := rt.payloadAttempts.Get(payload.ID)
+		if !ok {
+			attemptCount = 0
+		}
+
+		if attemptCount >= totalAttempt {
+			fmt.Printf("Payload %s has already been tried %d times. Skipping.\n", payload.ID, totalAttempt)
+			return nil
+		}
+
+		attemptCount++
+		rt.payloadAttempts.Set(payload.ID, attemptCount, util.DefaultCacheExpiration)
+
+		nextRunTime := time.Now().Add(retryDelay)
+		rt.nextRetries.Store(nextRunTime, payload)
+	} else {
+		return fmt.Errorf("Payload %s is not retryable. Skipping.\n", payload.ID)
+	}
+
+	return nil
+}
+
+// getterFn is a function that retrieves the retryTick for the given time.
+func (rt *retryTicker) getterFn(ctx context.Context, t time.Time) (Tick, error) {
+	rt.payloadAttempts.ClearExpired()
+
+	upkeepPayloads := []ocr2keepers.UpkeepPayload{}
+
+	rt.nextRetries.Range(func(key, value interface{}) bool {
+		if runTime, ok := key.(time.Time); ok {
+			if t.After(runTime) {
+				if payload, ok := value.(ocr2keepers.UpkeepPayload); ok {
+					upkeepPayloads = append(upkeepPayloads, payload)
+				}
+				rt.nextRetries.Delete(key)
+			}
+		}
+		return true
+	})
+
+	return retryTick{
+		upkeeps: upkeepPayloads,
+	}, nil
+}
+
+// NewRetryTicker creates a new retryTicker with the specified interval and observer.
+func NewRetryTicker(interval time.Duration, observer observer) *retryTicker {
+	rt := &retryTicker{
+		nextRetries:     sync.Map{},
+		payloadAttempts: util.NewCache[int](attemptsCacheExpiration),
+	}
+
+	rt.timeTicker = NewTimeTicker(interval, observer, rt.getterFn)
+
+	return rt
+}
+
+type retryTick struct {
+	upkeeps []ocr2keepers.UpkeepPayload
+}
+
+// GetUpkeeps returns the upkeeps contained in the retryTick.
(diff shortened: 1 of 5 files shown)


pkg/tickers/retry.go
```diff
@@ -0,0 +1,84 @@
...
+	nextRetries  map[time.Time]ocr2keepers.UpkeepPayload
+	newDurations map[string]time.Duration
+}
+
+// Retry adds a retryable result to the RetryTicker.
+func (rt *RetryTicker) Retry(result CheckResult) error {
+	payload := result.Payload
+	retryable := result.Retryable
+
+	if retryable {
```
EasterTheBunny, line 34: These can be combined into `if result.Retryable` without affecting code readability and it will be a minor simplification.

EasterTheBunny (changes requested):
Looks great! And it's exactly as we discussed. Just a few minor comments. Also, make sure to provide tests before merging, of course.

pkg/tickers/retry.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+// getterFn is a function that retrieves the retryTick for the given time.
+func (rt *RetryTicker) getterFn(ctx context.Context, t time.Time) (Tick, error) {
+	upkeepPayloads := []ocr2keepers.UpkeepPayload{}
+	for key, value := range rt.nextRetries {
+		if t.After(key) {
+			upkeepPayloads = append(upkeepPayloads, value)
+			delete(rt.nextRetries, key)
```
shileiwill, line 56: Are we removing this too early? What if the retry is not successful?

As discussed with @ferglor, we need to track the retry status somehow, and mark the Payload as retried_and_succeeded if it went through. 

@EasterTheBunny can shed some light here.

pkg/tickers/retry.go
shileiwill: if not retryable, throw an error

pkg/tickers/retry.go
```diff
@@ -0,0 +1,84 @@
...
+	if retryable {
+		newDuration, ok := rt.newDurations[payload.Hash]
+		if !ok {
+			newDuration = 1 * time.Second // First time for this payload, set initial duration
+		}
+
+		nextRunTime := time.Now().Add(newDuration)
+		rt.nextRetries[nextRunTime] = payload
+
+		newDuration *= 2 // Double the duration for the next retry
```
shileiwill, line 43: use exponential duration.

EasterTheBunny (approved):

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+	payload := result.Payload
+
+	if result.Retryable {
+		attemptCount, ok := rt.payloadAttempts.Get(payload.ID)
+		if !ok {
+			attemptCount = 0
+		}
+
+		if attemptCount >= totalAttempt {
+			fmt.Printf("Payload %s has already been tried %d times. Skipping.\n", payload.ID, totalAttempt)
```
ferglor, line 36: Should this be using `log` instead of `fmt`? If so, we don't need the new line at the end.

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+			return nil
+		}
+
+		attemptCount++
+		rt.payloadAttempts.Set(payload.ID, attemptCount, util.DefaultCacheExpiration)
+
+		nextRunTime := time.Now().Add(retryDelay)
+		rt.nextRetries.Store(nextRunTime, payload)
+	} else {
+		return fmt.Errorf("Payload %s is not retryable. Skipping.\n", payload.ID)
```
ferglor, line 46: Same here

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+		return true
+	})
+
+	return retryTick{
+		upkeeps: upkeepPayloads,
+	}, nil
+}
+
+// NewRetryTicker creates a new retryTicker with the specified interval and observer.
+func NewRetryTicker(interval time.Duration, observer observer) *retryTicker {
```
ferglor, line 76: Can we move this constructor to after the retryTicker declaration? 🙏

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+	timeTicker
+	nextRetries     sync.Map // time.Time -> ocr2keepers.UpkeepPayload
+	payloadAttempts *util.Cache[int]
+}
+
+// Retry adds a retryable result to the retryTicker.
+func (rt *retryTicker) Retry(result ocr2keepers.CheckResult) error {
+	payload := result.Payload
+
+	if result.Retryable {
```
ferglor, line 29: I wonder is it easier to read this function if we instead do:

```
if !result.Retryable {
  return fmt.Errorf("Payload %s is not retryable. Skipping.\n", payload.ID)
}

// do the retry stuff here
```

That way you get to drop the "else" block?

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+			attemptCount = 0
+		}
+
+		if attemptCount >= totalAttempt {
+			fmt.Printf("Payload %s has already been tried %d times. Skipping.\n", payload.ID, totalAttempt)
+			return nil
+		}
+
+		attemptCount++
+		rt.payloadAttempts.Set(payload.ID, attemptCount, util.DefaultCacheExpiration)
```
ferglor, line 41: Do we want the attempt count to expire after 5 seconds? I would have thought we would want to track total retry count over a much longer period?

pkg/tickers/retry.go
```diff
@@ -0,0 +1,94 @@
...
+
+		nextRunTime := time.Now().Add(retryDelay)
+		rt.nextRetries.Store(nextRunTime, payload)
```
ferglor, line 44: I'm not sure what happens under the hood here, but in a high volume environment is whatever this time resolves to when used as a key granular or unique enough to ensure that we won't overwrite payloads in the map? I would have thought using the payload ID as the key would have been safer?

pkg/tickers/retry.go
shileiwill: If we dont expire them, this map in memory will explode over time. IIUC, in one `session`, we keep track of only max 3 times of retry for a given payload. Once this payload expires, a new session starts, and we can retry max 3 times for the same payload.

pkg/tickers/retry.go
shileiwill: Good point. What if the nextRuntime collide? I am not sure the likelihood of collision. I guess you are right, at a given time, there should be just one payload in cache, so payload is a unique identifier. Would like to hear thoughts from @EasterTheBunny too.

pkg/tickers/retry.go
shileiwill: sure! what is your preferred way of ordering? variables, constructor and then others?

pkg/tickers/retry.go
shileiwill: Good callout.  I think `readability` really depends on people's preference and what have here is a very classic case. It is more natural for some people, including myself, to read code like "if true, do something, else do others". On the other hand, some people want to always short circuit early, "if something, return, then blah blah". If you feel strongly on this, I will make the suggested change.

PR #137, opened 2023-06-09
Encode and Decode for observation and outcome

shileiwill:


pkg/v3/ocrtypes.go
```diff
@@ -0,0 +1,39 @@
...
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+)
+
+type AutomationObservation struct {
+	Instructions []string
+	Metadata     map[string]interface{}
+	Performable  []ocr2keepers.CheckResult
+}
+
+func (a AutomationObservation) EncodeAutomationObservation() ([]byte, error) {
```
EasterTheBunny, line 15: Let's just call this `Encode` to keep naming simple.

pkg/v3/ocrtypes.go
```diff
@@ -0,0 +1,39 @@
...
+	return obs, err
+}
+
+type AutomationOutcome struct {
+	Instructions []string
+	Metadata     map[string]interface{}
+	Performable  []ocr2keepers.CheckResult
+}
+
+func (a AutomationOutcome) EncodeAutomationOutcome() ([]byte, error) {
```
EasterTheBunny, line 31: Same here. Go with `Encode`

pkg/v3/ocrtypes.go
```diff
@@ -0,0 +1,39 @@
+package ocrtypes
+
+import (
+	"encoding/json"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+)
+
+type AutomationObservation struct {
```
EasterTheBunny, line 9: I know we don't always do it, but try to start adding documentation to exported structs and functions that describes what they are or what they do.

In this case you would start with
```
// AutomationObservation ...
type AutomationObservation struct {
```

Go style of commenting structs and functions starts with the name of the struct or function and is followed with whatever else. There are automated documentation builders that collect these comments and build full documentation.

EasterTheBunny (changes requested):

EasterTheBunny (approved):

PR #139, opened 2023-06-13
[RFC]Observation fn for ocr3 plugin

shileiwill:


pkg/v3/ocr3.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Create new AutomationObservation
+	observation := AutomationObservation{
+		Instructions: automationOutcome.Instructions,
+		Metadata:     automationOutcome.Metadata,
+		Performable:  automationOutcome.Performable,
```
EasterTheBunny, line 45: These values should not be set using the outcome. You can just create a new observation and leave all fields blank.

pkg/v3/ocr3_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Create a sample outcome for decoding
+	outcome := ocr3types.OutcomeContext{
+		PreviousOutcome: []byte(`{"Instructions":["instruction1"],"Metadata":{"key":"value"},"Performable":[]}`),
+	}
+
+	// Define a mock hook function for testing pre-build hooks
+	mockPrebuildHook := func(outcome AutomationOutcome) error {
+		return nil
+	}
```
EasterTheBunny, line 25: In the test, you can do something here to ensure the prebuild hook runs.

pkg/v3/ocr3_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	// Add the mock pre-build hook to the plugin's PrebuildHooks
+	plugin.PrebuildHooks = append(plugin.PrebuildHooks, mockPrebuildHook)
+
+	// Define a mock build hook function for testing build hooks
+	mockBuildHook := func(observation *AutomationObservation, instructionStore InstructionStore, samplingStore SamplingStore, resultStore ResultStore) error {
+		return nil
+	}
```
EasterTheBunny, line 33: You can do something here that adds to the observation to ensure the build hook runs.

EasterTheBunny (changes requested):

pkg/v3/ocr3.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, err
+	}
+
+	// Execute pre-build hooks
+	for _, hook := range plugin.PrebuildHooks {
+		err := hook(automationOutcome)
+		if err != nil {
+			return nil, err
+		}
+	}
```
EasterTheBunny, line 39: I was just now thinking about the error handling here. It's similar to something I was working with today where the question is do you exit early on an error or run all hooks and returned a joined error? In my case, I went with the latter and returned a joined error after the for loop. That would indicate that the pre-build process failed but allows all parts to run.

pkg/v3/ocr3.go
```diff
@@ -0,0 +1,60 @@
...
+	// Create new AutomationObservation
+	observation := AutomationObservation{}
+
+	// Execute build hooks
+	for _, hook := range plugin.BuildHooks {
+		err := hook(&observation, plugin.InstructionSource, plugin.MetadataSource, plugin.ResultSource)
+		if err != nil {
+			return nil, err
+		}
+	}
```
EasterTheBunny, line 50: Same error handling here

shileiwill: discussed offline, this will be as is, since this is stateless and we want to short circuit.

EasterTheBunny (approved):

PR #140, opened 2023-06-14
Upkeeps result store

amirylm:
Key/value store for check results of upkeeps.

**Features**

* thread safe using map[string] and rw mutex
* time based eviction of items (upon intervals)
* notifications upon time eviction or removal of items
* view options for filtering, sorting and limiting


pkg/resultstore/store.go
```diff
@@ -0,0 +1,189 @@
...
+
+// TODO: make these configurable?
+var (
+	notifyQBufferSize = 128
+	storeTTL          = time.Minute
+	gcInterval        = time.Minute * 5
+)
+
+// element is an internal representation of a result.
+type element struct {
```
ferglor, line 21: Why not call this `result`?

pkg/resultstore/store.go
```diff
@@ -0,0 +1,189 @@
...
+		ctx, cancel := context.WithCancel(pctx)
+		defer cancel()
+
+		ticker := time.NewTicker(gcInterval)
+		for {
+			select {
+			case <-ticker.C:
+				s.gc()
+			case <-ctx.Done():
+				s.lggr.Println("result store context done, stopping gc")
```
ferglor, line 59: Should you call `ticker.Stop()` here? Or defer a call to it on line 54?  Otherwise we might have a memory leak?

pkg/resultstore/store.go
```diff
@@ -0,0 +1,189 @@
...
+// Add adds element/s to the store.
+func (s *resultStore) Add(results ...ocr2keepers.CheckResult) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	added := 0
+	for _, result := range results {
+		id := result.Payload.ID
+		_, ok := s.data[id]
+		if ok {
```
ferglor, line 82: Is it cleaner if we test for `!ok`, so that you add the data within this if block? That way you can eliminate the call to `continue`?

pkg/resultstore/store.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// View returns a copy of the data in the store.
+// It accepts filters that can be used to prepare the results view.
+// NOTE: we apply filters while holding the read lock, these functions must not block.
+func (s *resultStore) View(opts ...ocr2keepers.ViewOpt) ([]ocr2keepers.CheckResult, error) {
+	filters, comparators, limit := ocr2keepers.ViewOpts(opts).Apply()
+
+	var results []ocr2keepers.CheckResult
+	s.lock.RLock()
```
ferglor, line 108: Does it make sense to extract lines 108-132 into a function? That way you can lock and defer the unlock for just this section of code, rather than having to specify where the unlocking should happen?

pkg/resultstore/store.go
```diff
@@ -0,0 +1,189 @@
...
+// notify writes to the notifications channel.
+// NOTE: we drop notifications in case the channel is full
+func (s *resultStore) notify(op ocr2keepers.NotifyOp, data ocr2keepers.CheckResult) {
+	select {
+	case s.notifications <- ocr2keepers.Notification{
+		Op:   op,
+		Data: data,
+	}:
+	default:
+		s.lggr.Println("q is full, dropping result")
```
ferglor, line 176: `queue is full, dropping result` 🏄

pkg/stores.go
```diff
@@ -1,41 +1,76 @@
...
-import "fmt"
+// NotifyOp is an operation that can be notified by the ResultStore
+type NotifyOp uint8
 
-type ResultStore[T any] interface {
-	Add(T)
-	Remove(T)
-	View() ([]T, error)
+const (
+	NotifyOpNil NotifyOp = iota
```
ferglor, line 7: This isn't about your PR, but I just wanted to say that I hate iotas 😂 I think they're a pain to debug with! But I get that we use them in lots of places in chainlink so I think it's fine to keep this as-is! 👍

pkg/stores.go
```diff
@@ -1,41 +1,76 @@
...
 }
 
-func (s *resultStore[T]) View() ([]T, error) {
-	var result []T
-	for _, r := range s.data {
-		result = append(result, r)
+// WithOrderView enables to limit the amount of results that should be returned
+func WithLimit(limit int) ViewOpt {
+	return func(opts *viewOpts) {
+		opts.limit = limit
```
ferglor, line 74: Do all of the interfaces and types declared in this file need to be exported? Or can some of them be unexported? 🏄

ferglor (commented):
Just a few comments, but looks good! I think stopping the ticker might be the only blocking comment? 🙌

ferglor (approved):
Thanks for addressing the comments! 🙌

pkg/stores.go
amirylm: We can maybe place some in the `./resultstore` package

EasterTheBunny (approved):

PR #142, opened 2023-06-14
Recover lost tests

ferglor:
In this PR, we're adding tests back into the project that were previously lost as part of the refactor merge.

### DIFF
--- a/pkg/observation_test.go
+++ b/pkg/observation_test.go
@@ -1,4 +1,4 @@
-package ocr2keepers
+package ocr2keepers_test
 
 import (
 	"encoding/json"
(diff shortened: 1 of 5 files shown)


pkg/observation_test.go
```diff
@@ -1,4 +1,4 @@
-package ocr2keepers
+package ocr2keepers_test
```
EasterTheBunny, line 1: Why use a different package name for the tests here?

EasterTheBunny (approved):

pkg/observation_test.go
ferglor: There was an import cycle and it was easier to rename the test package to side step it rather than move around implementation packages 👍

PR #145, opened 2023-06-15
update retry logic to match product needs

EasterTheBunny:


pkg/v3/ocr3_test.go
```diff
@@ -1,4 +1,4 @@
-package v3
+package ocr2keepers
```
shileiwill, line 1: good catch, this is fixed in another PR thanks to Fergal. You should see it when you rebase.

pkg/v3/tickers/time.go
```diff
@@ -57,13 +57,14 @@ func (t *timeTicker) Start(ctx context.Context) error {
 
 	t.running.Store(true)
 
-	for tm := range t.ticker.C {
-		ctx, cancelFn := context.WithTimeout(ctx, t.interval)
+	ctx, cancel := context.WithCancel(ctx)
 
+Loop:
```
shileiwill, line 62: is this right indention? what if we dont have Loop, we can still break(L79) from the for loop in L63?

shileiwill (approved):

pkg/v3/tickers/retry.go
```diff
@@ -11,51 +11,67 @@ import (
...
+	ErrRetryDurationExceeded = fmt.Errorf("payload has exceed allowed retry window")
 )
 
 const (
-	retryDelay              = 500 * time.Millisecond
-	totalAttempt            = 3
-	attemptsCacheExpiration = 10 * retryDelay // 5 seconds
+	DefaultRetryDelay           = 10 * time.Second
+	DefaultMaxRetryDuration     = 5 * time.Minute
+	DefaultRetryCacheExpiration = 2 * DefaultMaxRetryDuration // 10 minutes
```
shileiwill, line 21: I guess you dont need to expose those?

pkg/v3/tickers/time.go
EasterTheBunny: Haha! It's weird, I know. This is the Go way of breaking from outer loops. Yup, it's the right indentation.

pkg/v3/tickers/retry.go
EasterTheBunny: I'm exposing them to indicate what they are in documentation. Exported constants are included in pre-built docs and doing this offers more clarity on what is behind the scenes. ¯\_(ツ)_/¯

PR #146, opened 2023-06-15
Log Trigger Pre-build Hook w/ Skeleton Plugin

EasterTheBunny:
To begin tying more pieces together, a new function was added to the log trigger eligibility flow that allows it to function as a pre-build hook for the plugin. This leads to more complex component stacking up to the plugin instance.

The result store was moved into the v3 package.

The plugin code was moved to a plugin package to avoid circular dependencies and to isolate plugin specific code.


pkg/v3/plugin/ocr3.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
+	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
+
+	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
+)
+
+type InstructionStore interface{}
+
+type SamplingStore interface{}
```
ferglor, line 15: Can these two interfaces be unexported?

ferglor (approved):

pkg/v3/flows/logtrigger.go
```diff
@@ -30,8 +30,6 @@ type PreProcessor interface {
 //go:generate mockery --name ResultStore --structname MockResultStore --srcpkg "github.com/smartcontractkit/ocr2keepers/pkg/v3/flows" --case underscore --filename resultstore.generated.go
 type ResultStore interface {
 	Add(...ocr2keepers.CheckResult)
```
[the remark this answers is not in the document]
EasterTheBunny: I think we should make a cleanup ticket to organize our interfaces at the end of our first pass of development. They are a bit scattered, which is ok for now.

pkg/v3/plugin/ocr3.go
EasterTheBunny: I'll do that.

EasterTheBunny: Actually, since they are defined in the build hooks, they are needed for now, right?

EasterTheBunny: I think they'll need to be moved to the same place as the result store interface eventually.

PR #147, opened 2023-06-16
pre-build hook that removes outcome results from staging

EasterTheBunny:


pkg/v3/hooks/prebuild.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
+)
+
+type resultRemover interface {
+	Remove(...string)
+}
+
+var PrebuildHookRemoveFromStaging = func(store resultRemover) func(ocr2keepersv3.AutomationOutcome) error {
```
ferglor, line 11: Why does this return a function instead of just exposing the functionality directly?

EasterTheBunny: The function signature for the pre-build hooks is `func(ocr2keepersv3.AutomationOutcome) error`, but a dependency for this hook is the result store. To bring the result store in scope, either we could make a struct with a single property (the result store) and have a member function that references the store, or create a function that returns a function with the result store in scope. ¯\_(ツ)_/¯

ferglor: I think a struct with a member function is less confusing but happy to go with whatever you prefer 👍

EasterTheBunny: A struct is more consistent with the rest of the project.

pkg/v3/hooks/prebuild.go
```diff
@@ -0,0 +1,29 @@
...
+
+type PrebuildHookRemoveFromStaging struct {
+	remover resultRemover
+}
+
+func (hook *PrebuildHookRemoveFromStaging) RunHook(outcome ocr2keepersv3.AutomationOutcome) error {
+	toRemove := make([]string, 0, len(outcome.Performable))
+
+	for _, result := range outcome.Performable {
+		toRemove = append(toRemove, result.Payload.ID)
```
ferglor, line 23: I wonder what the performance difference would be in calling `hook.remover.Remove(result.Payload.ID)` directly instead of building a slice of IDs to be removed? 🤔

ferglor (approved):

pkg/v3/hooks/prebuild.go
EasterTheBunny: That's a good question. That may be resolved in the result store. Maybe using multiple goroutines to remove values? Right now it just runs a for loop on the array of strings. So passing one string at a time just starts up a for loop each time with a single value. ¯\_(ツ)_/¯

PR #149, opened 2023-06-19
Add a block ticker that sends block histories on a channel

ferglor:
In this PR, we're adding a block ticker that receives block histories from a block subscriber, and attempts to send the block history into its own channel, when available. If the ticker's channel has already received a block history, then values received from the block subscriber are discarded until the ticker's channel can be written to again.

# Buffering update

In the latest version of the code, the ticker has been updated so that:

- If the input channel (`t.ch`) can provide a value and if the output channel (`t.C`) can receive a value, the value from `t.ch` is sent into `t.C`
- If the input channel (`t.ch`) can provide a value and if the output channel (`t.C`) cannot receive a value, the value from `t.ch` is written to `t.bufferedValue`
- If the input channel (`t.ch`) cannot provide a value and if the output channel (`t.C`) can receive a value and if `t.bufferedValue` is not nil, then `t.bufferedValue` is sent into `t.C`

### DIFF
--- a/pkg/v3/tickers/block.go
+++ b/pkg/v3/tickers/block.go
@@ -23,8 +23,7 @@ type blockTicker struct {
 	ch         chan ocr2keepers.BlockHistory
 	subscriber blockSubscriber
 	closer     sync.Once
-	ctx        context.Context
-	cancel     context.CancelFunc
+	stopCh     chan int
 }
 
 func NewBlockTicker(subscriber blockSubscriber) (*blockTicker, error) {
@@ -33,16 +32,13 @@ func NewBlockTicker(subscriber blockSubscriber) (*blockTicker, error) {
 		return nil, err
 	}
 
-	ctx, cancel := context.WithCancel(context.Background())
-
 	return &blockTicker{
 		chID:       chID,
 		ch:         ch,
 		C:          make(chan ocr2keepers.BlockHistory),
 		subscriber: subscriber,
 		closer:     sync.Once{},
-		ctx:        ctx,
-		cancel:     cancel,
+		stopCh:     make(chan int),
 	}, nil
 }
 
@@ -59,6 +55,8 @@ loop:
 		case <-ctx.Done():
 			err = ctx.Err()
 			break loop
+		case <-t.stopCh:
+			return nil
 		}
 	}
 	return err
@@ -66,9 +64,9 @@ loop:
 
 func (t *blockTicker) Close() {
 	t.closer.Do(func() {
-		t.cancel()
+		t.stopCh <- 1
 		if err := t.subscriber.Unsubscribe(t.chID); err != nil {
-			log.Fatalf("error unsubscribing: %v", err)
+			log.Printf("error unsubscribing: %v", err)
 		}
 	})
 }
--- a/pkg/v3/tickers/block.go
+++ b/pkg/v3/tickers/block.go
@@ -1,0 +1,72 @@
+package tickers
+
+import (
+	"context"
+	"log"
+	"sync"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+)
+
+type blockSubscriber interface {
+	// Subscribe provides an identifier integer, a new channel, and potentially an error
+	Subscribe() (int, chan ocr2keepers.BlockHistory, error)
+	// Unsubscribe requires an identifier integer and indicates the provided channel should be closed
+	Unsubscribe(int) error
+}
+
+// blockTicker is a struct that follows the same design paradigm as a time ticker but provides blocks
+// instead of time
+type blockTicker struct {
+	C          chan ocr2keepers.BlockHistory
+	chID       int
+	ch         chan ocr2keepers.BlockHistory
+	subscriber blockSubscriber
+	closer     sync.Once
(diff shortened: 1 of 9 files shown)


pkg/v3/tickers/block.go
```diff
@@ -0,0 +1,74 @@
...
+	chID       int
+	ch         chan ocr2keepers.BlockHistory
+	subscriber blockSubscriber
+	closer     sync.Once
+	ctx        context.Context
+	cancel     context.CancelFunc
+}
+
+func NewBlockTicker(subscriber blockSubscriber) (*blockTicker, error) {
+	chID, ch, err := subscriber.Subscribe()
```
EasterTheBunny, line 31: I think it would make more sense to put the subscribe in the Start method. Then the `Once` functionality would not be needed on close.

pkg/v3/tickers/block.go
```diff
@@ -0,0 +1,74 @@
...
+// blockTicker is a struct that follows the same design paradigm as a time ticker but provides blocks
+// instead of time
+type blockTicker struct {
+	C          chan ocr2keepers.BlockHistory
+	chID       int
+	ch         chan ocr2keepers.BlockHistory
+	subscriber blockSubscriber
+	closer     sync.Once
+	ctx        context.Context
+	cancel     context.CancelFunc
```
EasterTheBunny, line 27: Do we need these? If the Start method takes the run context the constructor cannot be the context provider.

pkg/v3/tickers/block.go
```diff
@@ -0,0 +1,74 @@
...
+		select {
+		case blockHistory := <-t.ch:
+			select {
+			case t.C <- blockHistory:
+				log.Print("forwarded")
+			default:
+			}
+		case <-ctx.Done():
+			err = ctx.Err()
+			break loop
```
EasterTheBunny, line 61: Maybe just return an error instead of breaking the loop?

pkg/v3/tickers/block.go
```diff
@@ -0,0 +1,74 @@
...
+	}, nil
+}
+
+func (t *blockTicker) Start(ctx context.Context) (err error) {
+loop:
+	for {
+		select {
+		case blockHistory := <-t.ch:
+			select {
+			case t.C <- blockHistory:
```
EasterTheBunny, line 55: Is there a way we can store the last block history such that when the reading channel opens, we can read from the last instead of waiting for the next?

EasterTheBunny (changes requested):

pkg/v3/tickers/block.go
ferglor: I did it this way so that the channel ID would be available to the instance as soon as client had created an instance with this constructor, since the Close() method passes the channel ID to the unsubscribe?

pkg/v3/tickers/block.go
ferglor: So when t.C starts to reach from t.ch, you want t.C to immediately fetch whatever is in t.ch? So make t.ch a buffered channel of length 1?

pkg/v3/tickers/block.go
ferglor: Oh this just adds a line of test coverage that we otherwise can't get to!

pkg/v3/tickers/block.go
EasterTheBunny: Ahhh. That makes sense. It's the same challenge with the context.

pkg/v3/tickers/block.go
EasterTheBunny: Kind of. I want the most recent item from `t.ch` to be pushed to `t.C` whenever `t.C` unblocks. So if the cadence of `t.C` reads is 4 seconds and the cadence of `t.ch` writes is 1.5 seconds, `t.ch` will write 2 times (3s) before `t.C` reads again. However, I don't want `t.C` to have to wait the extra 0.5 seconds for `t.ch` to write again.

If this is more complicated than it is worth to save 0.5 seconds, then maybe not bother. I'm just thinking of slower chains with 10+seconds on block cadence. It feels convoluted and maybe we can come back around to it.

ferglor: So like, am I right in thinking that t.ch should manage its own internal buffer, e.g.:

- If t.ch can send the current value to t.C, then send it
- If t.C is blocked, and t.ch wants to send a value, t.ch should overwrite an e.g. `nextValueBuffer` field, continuously for as long as t.C is blocked
- When t.C unblocks:
  - If t.ch has a value, then send it, and wipe `nextValueBuffer`
  - If t.ch doesn't have a value, then try to read from `nextValueBuffer`, if the value is set, then send it, otherwise do nothing until t.ch has something to send

ferglor: Or we could flip it so that the ticker always pulls into a temporary buffer and tries to send the value itself when t.ch has got nothing to send?

ferglor: Oh wait I'm not sure if we _can_ flip it, it might need to be buffered by t.ch?

EasterTheBunny: This looks right to me

EasterTheBunny (approved):

PR #150, opened 2023-06-19
Log observer: pre-processing base types

amirylm:
Updating `UpkeepPayload` to include all needed fields for the log event provider to construct the payload for log upkeeps.


EasterTheBunny (approved):
LGTM

PR #152, opened 2023-06-19
Add a metadata store

ferglor:
In this PR, we're renaming the SamplingStore to the MetadataStore, and implementing the respective functionality. When the MetadataStore runs, we listen to a block ticker and retrieve BlockHistory slices. The BlockHistory slice overwrites the locally stored BlockHistory slice. A set function has been added that lets us store a slice of upkeep IDs.

### DIFF
--- a/pkg/v3/stores.go
+++ b/pkg/v3/stores.go
@@ -1,7 +1,10 @@
 package ocr2keepers
 
 import (
+	"sync"
+
 	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/tickers"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/v3/stores.go
```diff
@@ -78,3 +90,59 @@ func WithLimit(limit int) ViewOpt {
 		opts.limit = limit
 	}
 }
+
+type metadataStore struct {
+	ticker       *tickers.BlockTicker
```
EasterTheBunny, line 95: I think it is fine for now, but I think we shouldn't conflate the store with tickers. Just to maintain consistency, tickers should only be connected to an observer to create observation flows instead.

EasterTheBunny (approved):

PR #155, opened 2023-06-20
fix tests

EasterTheBunny:


pkg/v3/plugin/ocr3.go
```diff
@@ -7,19 +7,27 @@ import (
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
 	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
 	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
 )
 
 type InstructionStore interface{}
 
 type SamplingStore interface{}
```
shileiwill, line 16: remove

pkg/v3/plugin/ocr3.go
```diff
@@ -7,19 +7,27 @@ import (
...
+}
+
 type ocr3Plugin[RI any] struct {
-	PrebuildHooks     []func(ocr2keepersv3.AutomationOutcome) error
-	BuildHooks        []func(*ocr2keepersv3.AutomationObservation, InstructionStore, SamplingStore, ocr2keepersv3.ResultStore) error
-	InstructionSource InstructionStore
-	MetadataSource    SamplingStore
-	ResultSource      ocr2keepersv3.ResultStore
+	PrebuildHooks      []func(ocr2keepersv3.AutomationOutcome) error
+	BuildHooks         []func(*ocr2keepersv3.AutomationObservation, InstructionStore, SamplingStore, ocr2keepersv3.ResultStore) error
```
shileiwill, line 24: v3.InstructionStore etc

pkg/v3/plugin/ocr3.go
```diff
@@ -70,8 +78,40 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
 	panic("ocr3 Outcome not implemented")
 }
 
-func (plugin *ocr3Plugin[RI]) Reports(seqNr uint64, outcome ocr3types.Outcome) ([]ocr3types.ReportWithInfo[RI], error) {
-	panic("ocr3 Reports not implemented")
+func (plugin *ocr3Plugin[RI]) Reports(seqNr uint64, raw ocr3types.Outcome) ([]ocr3types.ReportWithInfo[RI], error) {
```
shileiwill, line 81: seqNr not used.

pkg/v3/plugin/ocr3.go
```diff
@@ -70,8 +78,40 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
...
+
+	if outcome, err = ocr2keepersv3.DecodeAutomationOutcome(raw); err != nil {
+		return nil, err
+	}
+
+	toPerform := []ocr2keepers.CheckResult{}
+	var gasUsed uint64
+
+	for i, result := range outcome.Performable {
+		if gasUsed+result.GasUsed > plugin.ReportGasLimit || len(toPerform) > plugin.MaxUpkeepBatchSize {
```
shileiwill, line 96: Just to confirm, is it possible that in one call of this Reports function, gasUsed and len(toPerform) never satisfy the above two checks, so it will never come to this if hence reports not populated properly.

pkg/v3/plugin/ocr3.go
```diff
@@ -70,8 +78,40 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
...
+			encoded, encodeErr := plugin.ReportEncoder.Encode(toPerform...)
+			err = errors.Join(err, encodeErr)
+
+			reports = append(reports, ocr3types.ReportWithInfo[RI]{
+				Report: types.Report(encoded),
```
shileiwill, line 101: if `encodeErr` is not nil, what we are getting for `encoded`? should we add it to `reports`?

pkg/v3/plugin/ocr3.go
```diff
@@ -70,8 +78,40 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
...
+			})
+
+			toPerform = []ocr2keepers.CheckResult{}
+			gasUsed = 0
+
+			continue
```
shileiwill, line 107: i guess this is the expected behavior, we skip L110-111 and start the next report/toPerform

pkg/v3/plugin/plugin.go
```diff
@@ -38,7 +39,10 @@ func New(
 		PrebuildHooks: []func(ocr2keepersv3.AutomationOutcome) error{
 			ltFlow.ProcessOutcome,
 		},
-		ResultSource: rs,
+		ResultSource:       rs,
+		ReportEncoder:      encoder,
+		ReportGasLimit:     5_000_000,
+		MaxUpkeepBatchSize: 1,
```
shileiwill, line 45: Am i understanding this right?

We have 2 `outcome.Performable` as a batch and the 2 lives in a report, as indicated in L96 of ocr3.go.

shileiwill (approved):
LGTM

pkg/v3/plugin/ocr3.go
EasterTheBunny: remove the types? I think this is almost ready to happen with some other PRs coming through.

pkg/v3/plugin/ocr3.go
EasterTheBunny: This should be part of another PR coming through.

pkg/v3/plugin/ocr3.go
EasterTheBunny: I'll use an `_` instead.

pkg/v3/plugin/ocr3.go
EasterTheBunny: Good catch

pkg/v3/plugin/ocr3.go
EasterTheBunny: It shouldn't be added to the report. I'll change that.

pkg/v3/plugin/ocr3.go
EasterTheBunny: This was a mistake of mine. The intent is to add to a report and clear the tracking info but still add the current result to the tracking.

pkg/v3/plugin/plugin.go
EasterTheBunny: Yes. These are just hard-coded values for now, but will be set by configurations going forward.

PR #156, opened 2023-06-20
Outcome generation

ferglor:
In this PR, we're implementing the `Outcome` function of the v3 plugin

In this function, we receive a list of `AttributedObservations`. We decode each observation, and we count the unique instances of each `PerformableResult` in the observation. 

Every performable result is comprised of three fields:

- Eligible (bool)
- Retryable (bool)
- Payload (UpkeepPayload)

However, only the `Payload` field is used to distinguish one performable result from another; or to put it another way, the Eligible and Retryable fields are ignored in the uniqueness comparison, and only the Payloads of two performable results are compared to determine if they are the same performable result or not.

Once the number of unique performable results has been calculated, we copy any result that exceeds the quorum threshold (`= number of attributed observations / 2`) into the outcome.

### DIFF
--- a/pkg/v3/plugin/ocr3.go
+++ b/pkg/v3/plugin/ocr3.go
@@ -83,7 +83,7 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
 		}
 
 		for _, result := range observation.Performable {
-			uid := fmt.Sprintf("%v", result.Payload)
+			uid := fmt.Sprintf("%v", result)
 			payloadCount, ok := resultCount[uid]
 			if !ok {
 				payloadCount = resultAndCount{
--- a/pkg/v3/plugin/ocr3.go
+++ b/pkg/v3/plugin/ocr3.go
@@ -12,8 +12,6 @@ import (
 	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
 )
 
-var ErrNotEnoughInputs = fmt.Errorf("not enough inputs")
-
 type InstructionStore interface{}
 
 type SamplingStore interface{}
--- a/pkg/v3/plugin/ocr3.go
+++ b/pkg/v3/plugin/ocr3.go
@@ -74,7 +74,6 @@ func (plugin *ocr3Plugin[RI]) Outcome(outctx ocr3types.OutcomeContext, query typ
 		count  int
 	}
 
-	submittedObservations := len(attributedObservations)
 	resultCount := make(map[string]resultAndCount)
 
 	for _, attributedObservation := range attributedObservations {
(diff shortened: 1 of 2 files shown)


pkg/v3/plugin/ocr3.go
```diff
@@ -66,8 +68,50 @@ func (plugin *ocr3Plugin[RI]) ValidateObservation(outctx ocr3types.OutcomeContex
...
+	resultCount := make(map[string]resultAndCount)
+
+	for _, attributedObservation := range attributedObservations {
+		observation, err := ocr2keepersv3.DecodeAutomationObservation(attributedObservation.Observation)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, result := range observation.Performable {
+			uid := fmt.Sprintf("%v", result.Payload)
```
EasterTheBunny, line 86: Ideally the entire `CheckResult` should be identical.

pkg/v3/plugin/ocr3.go
```diff
@@ -66,8 +68,50 @@ func (plugin *ocr3Plugin[RI]) ValidateObservation(outctx ocr3types.OutcomeContex
...
+				}
+			} else {
+				payloadCount.count++
+			}
+			resultCount[uid] = payloadCount
+		}
+	}
+
+	submittedObservations := len(attributedObservations)
+	quorumThreshold := submittedObservations / 2
```
EasterTheBunny, line 101: I think this is fine for now, but maybe we should set a lower threshold too. That is to say, the quorum threshold should not be below (f+1). More observations pushes the threshold higher, but fewer moves it to the OCR quorum as a limiter.

EasterTheBunny (approved):

pkg/v3/plugin/ocr3.go
ferglor: I wasn't sure if retryable would vary across different nodes, as in, for one node the check was successful the first time, but on another node the same check failed initially but was successful afterwards?

EasterTheBunny: Retryable is reset for every run through the check pipeline. In fact, the check pipeline only receives an UpkeepPayload so it doesn't know anything about a previous run. I think the only deviation between these check results is in the specific values of fastgas and linknative prices. But we rely on all of it to be identical anyway in the OCR2 version and it's working fine there.

pkg/v3/plugin/ocr3.go
EasterTheBunny: We can have access to `f` from the off-chain config. But now that I think about it, OCR won't produce an Outcome anyway if it doesn't have an `f+1` quorum. So even if the threshold is below `f+1`, an Outcome can be produced by a single node but the network won't be able to create one.

Therefore, what you have should be fine.

PR #157, opened 2023-06-20
Basic plugin factory

EasterTheBunny:


ferglor (approved):
Looks good but all the for loops can be replaced with:

for _, svc := range svcs

To prevent you having to index the slice

EasterTheBunny:
> 

Since the values are sent to a separate goroutine, the value goes out of scope right? I've had that trouble before with ranges where I have had to reference by index to the original instead of use the temporary value provided by the range call.

PR #165, opened 2023-06-21
set block in recovery ticker

shileiwill:


pkg/v3/tickers/recovery.go
```diff
@@ -1,19 +1,28 @@
...
 type recoveryTicker struct {
 	*retryTicker
-	// lastBlock ocr2keepers.BlockKey
+	lock      sync.RWMutex
+	lastBlock ocr2keepers.BlockKey
 }
 
-// getterFn is a function that retrieves the recoveryTick for the given time.
+// setBlock updates the block of the given payload
+func (rt *recoveryTicker) setBlock(p ocr2keepers.UpkeepPayload) {
```
ferglor, line 17: Won't `p` need to be a pointer to update the `CheckBlock` here?

pkg/v3/tickers/recovery.go
```diff
@@ -1,19 +1,28 @@
...
+	rt.lock.Lock()
+	defer rt.lock.Unlock()
```
ferglor, line 19: I think this can be:

```
rt.lock.RLock()
defer rt.lock.RUnlock()
```

Since the `rt` is being read from, not written to?

pkg/v3/tickers/recovery.go
```diff
@@ -1,19 +1,28 @@
...
+	p.CheckBlock = rt.lastBlock
+}
+
+// payloadModFunc is a function that updates the payload
 func (rt *recoveryTicker) payloadModFunc(p ocr2keepers.UpkeepPayload) ocr2keepers.UpkeepPayload {
```
ferglor, line 24: Could this function be renamed to something like `modifyPayload`?

ferglor (commented):
Can we have some unit tests for these functions? 🙏

pkg/v3/tickers/recovery.go
EasterTheBunny, line 17: The goal of this function is to be exported on the ticker such that an outside component can set the block. Something like `SetBlock(UpkeepKey)`. It is not intended to be used by the ticker itself.

pkg/v3/tickers/recovery.go
EasterTheBunny: This is correct because this function should write to the `latestBlock`

pkg/v3/tickers/recovery.go
```diff
@@ -1,19 +1,28 @@
...
-	// TODO: update block in payload
+	rt.setBlock(p)
```
EasterTheBunny, line 25: We don't need to use this function here. You can just update the upkeep payload block number directly and then return the modified payload. Ensure you use RLock for this.

EasterTheBunny (changes requested):

pkg/v3/tickers/recovery.go
ferglor: Ah okay, so the logic on line 20 should be flipped?

pkg/v3/tickers/recovery.go
EasterTheBunny: That makes sense to me.

EasterTheBunny (approved):
LGTM

PR #166, opened 2023-06-22
Log observer: pre processing

amirylm:


pkg/v3/flows/logtrigger.go
```diff
@@ -107,13 +113,18 @@ func newRetryFlow(rs ResultStore, rn ocr2keepersv3.Runner, recoverer Retryer, co
...
+}
 
-func (et emptyTick) GetUpkeeps(context.Context) ([]ocr2keepers.UpkeepPayload, error) {
-	return nil, nil
+func (et logTick) GetUpkeeps(ctx context.Context) ([]ocr2keepers.UpkeepPayload, error) {
+	if et.logProvider == nil {
+		return nil, nil
+	}
+	return et.logProvider.GetLogs(ctx)
 }
```
EasterTheBunny, line 125: I like this. It distinguishes the difference between a basic time tick and the log tick source.

pkg/v3/flows/logtrigger_test.go
```diff
@@ -28,7 +28,7 @@ func TestLogTriggerEligibilityFlow_EmptySet(t *testing.T) {
 	// values every time
 	src.On("PreProcess", mock.Anything, mock.Anything).Return([]ocr2keepers.UpkeepPayload{}, nil).Times(2)
 
-	_, svcs := NewLogTriggerEligibility(src, store, runner, logger)
+	_, svcs := NewLogTriggerEligibility(src, store, runner, nil, logger)
```
EasterTheBunny, line 31: Here it would be best to test the log provider as a dependency instead of the preprocess. The `src` mock should probably mock the log provider and ensure the appropriate function is called. The number of calls should end up being the same.

EasterTheBunny (changes requested):
Looks like what I had in mind. Just a few changes to remove things that won't be needed anymore and test cleanup.

EasterTheBunny (approved):

PR #167, opened 2023-06-22
Add coordinate block prebuild and build hooks

ferglor:
In this PR we're:

- Restructuring hooks into`prebuild` and `build` packages
- Moving the instruction store and instructions into their own package (we were running into import cycles in tests)
- Adding a new coordinate block prebuild hook that:
  -  Reads the instructions from the outcome, and if `ShouldCoordinateBlock` is present, adds `DoCoordinateBlock` to the instruction store
- Adding a new coordinate block build hook that:
  - reads the instruction store for the `ShouldCoordinateBlock` instruction, and if present, broadcasts the `ShouldCoordinateBlock` instruction on the observation
  - reads the instruction store for the `DoCoordinateBlock` instruction, and if present, adds the block history from the metadata store into the observation's metadata

### DIFF
--- a/pkg/v3/hooks/build/coordinate_block.go
+++ b/pkg/v3/hooks/build/coordinate_block.go
@@ -1,0 +1,30 @@
+package build
+
+import (
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg/v3"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/flows"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/instruction"
+)
+
+const (
+	ShouldCoordinateBlock instruction.Instruction = "should coordinate block"
+	DoCoordinateBlock     instruction.Instruction = "do coordinate block"
+)
+
+type BuildHook func(*ocr2keepers.AutomationObservation, instruction.InstructionStore, ocr2keepers.MetadataStore, flows.ResultStore) error
+
+type coordinateBlockHook struct{}
+
+func NewCoordinateBlockHook() *coordinateBlockHook {
+	return &coordinateBlockHook{}
+}
+
+func (h *coordinateBlockHook) RunHook(obs *ocr2keepers.AutomationObservation, instructionStore instruction.InstructionStore, metadataStore ocr2keepers.MetadataStore, resultStore flows.ResultStore) error {
+	if instructionStore.Has(ShouldCoordinateBlock) {
+		obs.Instructions = append(obs.Instructions, ShouldCoordinateBlock)
+	} else if instructionStore.Has(DoCoordinateBlock) {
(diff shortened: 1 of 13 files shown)


pkg/v3/hooks/build/coordinate_block.go
```diff
@@ -0,0 +1,31 @@
...
+func NewCoordinateBlockHook() *coordinateBlockHook {
+	return &coordinateBlockHook{}
+}
+
+func (h *coordinateBlockHook) RunHook(obs *ocr2keepers.AutomationObservation, instructionStore instruction.InstructionStore, metadataStore ocr2keepers.MetadataStore, resultStore flows.ResultStore) error {
+	if instructionStore.Has(ShouldCoordinateBlock) {
+		obs.Instructions = append(obs.Instructions, ShouldCoordinateBlock)
+	} else if instructionStore.Has(DoCoordinateBlock) {
+		obs.Metadata["blockHistory"] = metadataStore.GetBlockHistory()
+		instructionStore.Delete(ShouldCoordinateBlock)
```
EasterTheBunny, line 27: I don't think we will get to this point if there is a should in the instruction store. The first if condition will trigger first. I think the pre-build hook should be responsible for removing the should and adding the do based on what is in the outcome. We probably shouldn't do writes to the stores within the build hooks to maintain a one-way data flow. This will also maintain state consistency across multiple build hooks.

EasterTheBunny (changes requested):

pkg/v3/hooks/prebuild/coordinate_block.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+}
+
+func (h *coordinateBlockHook) RunHook(outcome ocr2keepersv3.AutomationOutcome) error {
+	for _, instruction := range outcome.Instructions {
+		if instruction == ShouldCoordinateBlock {
+			h.instructionStore.Set(DoCoordinateBlock)
+			h.instructionStore.Delete(ShouldCoordinateBlock)
+			break
+		}
```
EasterTheBunny, line 29: I missed this one earlier, but we should also remove the do instruction when a coordinated block comes in the outcome. I don't think we have a key yet for the outcome metadata for coordinated block, but if that key exists, we should remove the do instruction and send the coordinated block to the coordinated block ticker. That ticker is in another PR, but you can just add a dependency for it in this run hook.

EasterTheBunny (changes requested):

pkg/v3/hooks/prebuild/coordinate_block.go
ferglor: Do you know what PR that ticker is in? The block ticker that was merged before was for block histories, but we need one for ocr2keepers.BlockKey, is that right?

EasterTheBunny (approved):
LGTM

PR #172, opened 2023-06-23
enable dynamic config for retry and recovery

shileiwill:
To support both retryConfig and recoveryConfig in the main flow(NewLogTriggerEligibility), I have to change the variadic arguments to 2 slices of arrays. The downside of this change is we can't skip the config(s) when invoking the function. Let me know if there is a better way.


pkg/v3/tickers/recovery.go
```diff
@@ -8,6 +8,18 @@ import (
 	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
 )
 
+const (
+	DefaultRecoveryDelay           = 1 * time.Minute
+	DefaultMaxRecoveryDuration     = 60 * time.Minute
```
RyanRHall, line 13: should we make these configurable through env vars? I'm nervous we'll want to try to change these at some point and if they're hard-coded it will have to be via a release.

PR #173, opened 2023-06-23
Auto 3298 sample ticker

EasterTheBunny:


pkg/v3/tickers/sample.go
```diff
@@ -0,0 +1,150 @@
...
+		select {
+		case h := <-st.blocks.C:
+			if len(h) == 0 {
+				continue
+			}
+
+			ctx, cancelFn := context.WithTimeout(ctx, st.samplingLimit)
+
+			// do the observation with limited time
+			tick, err := st.getterFn(ctx, h[0])
```
shileiwill, line 70: n00b: why we get only the first element?

pkg/v3/tickers/sample.go
```diff
@@ -0,0 +1,150 @@
...
+	)
+
+	// TODO: convert to block key ticker instead of time ticker to provide
+	// block scope to active upkeep provider
+	if upkeeps, err = ticker.getter.GetActiveUpkeeps(ctx, block); err != nil {
+		return nil, err
+	}
+
+	upkeeps = ticker.shuffler.Shuffle(upkeeps)
+	size := ticker.ratio.OfInt(len(upkeeps))
```
shileiwill, line 117: n00b, didn't get how this is used? it is said to be n out of x. x is the length of upkeeps, how about n?

pkg/v3/tickers/sample.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if len(upkeeps) == 0 || size <= 0 {
+		return upkeepTick{upkeeps: nil}, nil
+	}
+
+	return upkeepTick{upkeeps: upkeeps[:size]}, nil
+}
+
+func NewSampleTicker(
```
shileiwill, line 126: this function is not used or test covered.

shileiwill (approved):
Left a few minor comments for my learning. Approved to unblock.

pkg/v3/tickers/sample.go
EasterTheBunny: Since BlockHistory is an array of BlockKey, we need the latest BlockKey. Ideally we should create a helper function for this on BlockHistory. I'll do that.

pkg/v3/tickers/sample.go
EasterTheBunny: I completely missed the tests. Will add them now.

EasterTheBunny:
We talked about making the sample ticker time based. To keep from over-complicating this PR, we can merge this and change it to time based later. This will allow the sampling flow to be put together, at least since it builds on this work.

shileiwill (approved):

PR #178, opened 2023-07-13
[RFC] Draft Coordinator

shileiwill:
First readable version of log event coordinator. Please comment and validate if the logic makes sense. I owe unit tests to this PR, will add to this PR once we are aligned on the logic!


pkg/v3/coordinator/coordinator.go
```diff
@@ -46,16 +49,44 @@ func NewReportCoordinator(logs EventProvider, conf config.OffchainConfig, logger
...
 
+// This is called by the OCR plugin on ShouldAcceptFinalizedReport
+// When this upkeep comes, we put it in the cache(activeKeys) and mark it as true(upkeep confirmed by log poller?)
+
+// As in Doc: Stores the trigger, upkeepID and triggerID for a report which is inflight (called via plugin shouldAccept)
 func (rc *reportCoordinator) Accept(upkeep ocr2keepers.ReportedUpkeep) {
-	if _, ok := rc.activeKeys.Get(upkeep.ID); !ok {
-		rc.activeKeys.Set(upkeep.ID, true, util.DefaultCacheExpiration)
+	// Checking if Extension is a map and txHash exists
+	if extension, ok := upkeep.Trigger.Extension.(map[string]interface{}); ok {
```
shileiwill, line 58: is this how we get log trigger upkeeps?

pkg/v3/coordinator/log_event_coordinator.go
```diff
@@ -46,15 +46,38 @@ func NewReportCoordinator(logs EventProvider, conf config.OffchainConfig, logger
...
+	// Return true if all checks pass
+	return true
+}
+
 func (rc *reportCoordinator) Accept(upkeep ocr2keepers.ReportedUpkeep) {
+	isLogEvent := rc.isLogEventUpkeep(upkeep)
+	if !isLogEvent {
+		rc.logger.Printf("Upkeep is not log event based, skipping ID: %s", upkeep.ID)
+		return
+	}
```
EasterTheBunny, line 71: One note on coding style that isn't super important:

I would rewrite this as the following just to avoid introducing another variable into scope and it reads cleaner.
```
if !rc.isLogEventUpkeep(upkeep) {
    // ...
}
```

pkg/v3/coordinator/log_event_coordinator.go
```diff
@@ -102,6 +125,10 @@ func (rc *reportCoordinator) checkEvents(ctx context.Context) error {
 			rc.logger.Printf("Skipping perform log in transaction %s as confirmations (%d) is less than min confirmations (%d)", evt.TransactionHash, evt.Confirmations, rc.minConfs)
 			continue
 		}
+		if evt.Type == ocr2keepers.StaleReportEvent {
+			rc.logger.Printf("Skipping perform log for upkeep %s in transaction %s as it is stale", evt.ID, evt.TransactionHash)
+			continue
+		}
```
EasterTheBunny, line 131: Instead of skipping the stale log, we should add it into the cache to ensure the id continues getting blocked in filters.

EasterTheBunny (changes requested):
A couple of quick notes

pkg/v3/coordinator/log_event_coordinator.go
```diff
@@ -113,7 +140,7 @@ func (rc *reportCoordinator) performEvent(evt ocr2keepers.TransmitEvent) {
 	if confirmed, ok := rc.activeKeys.Get(evt.ID); ok {
```
shileiwill, line 140: I didnt really touch this performEvent func, is this good?

pkg/v3/plugin/ocr3.go
```diff
@@ -208,12 +212,14 @@ func (plugin *ocr3Plugin[RI]) ShouldTransmitAcceptedReport(_ context.Context, _
...
 
 	for _, upkeep := range upkeeps {
-		if !plugin.Coordinator.IsTransmissionConfirmed(upkeep) {
-			// if any upkeep in the report does not have confirmation, attempt
-			// again
-			return true, nil
+		for _, coord := range plugin.Coordinators {
+			if !coord.IsTransmissionConfirmed(upkeep) {
+				// if any upkeep in the report does not have confirmation, attempt again
+				// TODO: is this logic correct? We skip other upkeeps in the report and other coordinators?
```
EasterTheBunny, line 218: This might need to be if all are false. This is going to be delicate I think.

shileiwill: sorry @EasterTheBunny i didnt quite follow. do you mean we "return true, nil" only if all upkeeps are transmission confirmed(coord.IsTransmissionConfirmed(ALL upkeepS))?

EasterTheBunny: Instead of exiting early on a single coordinator returning false, collect all values and return true if all coordinators return false.

EasterTheBunny (approved):

PR #179, opened 2023-07-13
Time ticker observe limit

EasterTheBunny:
Allow observers to be configured with a context time limit on `Process`. This is
apart from the ticker interval and all tickers should not set a limit on
processing time. Now the observers will be responsible for limiting their
processing by configuration.


shileiwill (approved):
LGTM! I didnt see a test covering the timeout scenario, am i missing it?

EasterTheBunny:
> LGTM! I didnt see a test covering the timeout scenario, am i missing it?

Huh. You're right. I didn't. I did for the initialization ¯\_(ツ)_/¯

PR #180, opened 2023-07-13
support trigger id for upkeep

shileiwill:
As title, a thin helper function to get the trigger ID.

### DIFF
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -120,6 +120,29 @@ func (p UpkeepPayload) GenerateID() string {
 	return hex.EncodeToString(idh[:])
 }
 
+func (p UpkeepPayload) TriggerID() (string, error) {
+	var extension map[string]interface{}
+	var ok bool
+
+	if extension, ok = p.Trigger.Extension.(map[string]interface{}); !ok {
+		return "", fmt.Errorf("Failed to generate trigger ID for upkeep. Extension is not a map")
+	}
+
+	txHash, ok := extension["txHash"].(string)
+	if !ok {
+		return "", fmt.Errorf("Failed to generate trigger ID for upkeep. Tx hash does not exist in Extension")
+	}
+
+	logIndex, ok := extension["logIndex"].(string)
+	if !ok {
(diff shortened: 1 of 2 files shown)


pkg/basetypes.go
```diff
@@ -120,6 +120,29 @@ func (p UpkeepPayload) GenerateID() string {
 	return hex.EncodeToString(idh[:])
 }
 
+func (p UpkeepPayload) TriggerID() (string, error) {
+	var extension map[string]interface{}
+	var ok bool
+
+	if extension, ok = p.Trigger.Extension.(map[string]interface{}); !ok {
+		return "", fmt.Errorf("Failed to generate trigger ID for upkeep. Extension is not a map")
+	}
```
EasterTheBunny, line 129: The extension can be nil so testing for this might be too aggressive of a validity test.

pkg/basetypes.go
```diff
@@ -120,6 +120,29 @@ func (p UpkeepPayload) GenerateID() string {
...
+
+	txHash, ok := extension["txHash"].(string)
+	if !ok {
+		return "", fmt.Errorf("Failed to generate trigger ID for upkeep. Tx hash does not exist in Extension")
+	}
+
+	logIndex, ok := extension["logIndex"].(string)
+	if !ok {
+		return "", fmt.Errorf("Failed to generate trigger ID for upkeep. Log index does not exist in Extension")
+	}
```
EasterTheBunny, line 139: These are available only to log event triggers but we want conditional triggers too.

pkg/basetypes.go
shileiwill: I see. What is you recommended behavior when Extension is nil?

pkg/basetypes.go
shileiwill: My current proposal is to support only log events. If it is conditional, and it tries to fetch TriggerID, we give error saying that conditional doesnt have triggerID. 

I assume the above is not sound. Would like to hear the ideal way of dealing with conditionals.

PR #182, opened 2023-07-14
Plugin Generics

EasterTheBunny:
The plugin gets to set its own report info type. This commit defines that type and formats the plugin specifically for that info type.


shileiwill (approved):
LG!

PR #184, opened 2023-07-17
add insufficient fund event type and add consistent trigger id

shileiwill:


amirylm (approved):

pkg/basetypes.go
```diff
@@ -45,12 +46,13 @@ const (
 	PerformEvent TransmitEventType = iota
 	StaleReportEvent
 	ReorgReportEvent
+	InsufficientFundsReportEvent
```
EasterTheBunny, line 49: This event is emitted in the transmit function on the contract? I must have missed that one.

EasterTheBunny (approved):

pkg/basetypes.go
shileiwill: Yes, that is from the contract.

PR #186, opened 2023-07-18
Complete e2e with libocr Alpha Release

EasterTheBunny:
This commit completes a plugin factory and delegate using the latest alpha
release of libocr. Changes to the v3 simulator with no guarantee that it works
yet.


shileiwill (approved):
LGTM!

PR #187, opened 2023-07-18
add coordinator to preprocessor

shileiwill:


pkg/v3/observer.go
```diff
@@ -92,6 +94,8 @@ func (o *Observer[T, R]) Process(ctx context.Context, tick tickers.Tick[[]T]) er
 		return err
 	}
 
+	fmt.Println(len(results))
+	fmt.Println("------")
```
EasterTheBunny, line 98: These should be removed.

pkg/v3/plugin/ocr3.go
```diff
@@ -25,6 +25,7 @@ type Encoder interface {
 type Coordinator interface {
 	Accept(ocr2keepers.ReportedUpkeep) error
 	IsTransmissionConfirmed(ocr2keepers.ReportedUpkeep) bool
+	PreProcess(context.Context, []ocr2keepers.UpkeepPayload) ([]ocr2keepers.UpkeepPayload, error)
 }
```
EasterTheBunny, line 29: I see the reason for adding to this interface, but one thing to keep in mind is interface bloat. Because `Accept` and `IsTransmissionConfirmed` are related to different behavior than `PreProcess`, ideally they should be separate interfaces. That is, a coordinator should implement the first two, but the last one may not be for every coordinator. `PreProcess` is already defined in the `Preprocessor` interface.

pkg/v3/resultstore/store.go
```diff
@@ -82,6 +83,7 @@ func (s *resultStore) Notifications() <-chan ocr2keepersv3.Notification {
 
 // Add adds element/s to the store.
 func (s *resultStore) Add(results ...ocr2keepers.CheckResult) {
+	fmt.Print("welcome to result store add")
```
EasterTheBunny, line 86: This can also be removed.

EasterTheBunny (changes requested):

pkg/v3/flows/logtrigger_test.go
```diff
@@ -186,15 +207,16 @@ func TestLogTriggerEligibilityFlow_Retry(t *testing.T) {
 
 	// after the upkeep is determined to be eligible and not retryable, the
 	// result is added to the result store
-	rStore.On("Add", mock.Anything).Times(1)
+	rStore.On("Add", mock.Anything).Times(4)
 
 	// metadata store should set the value thrice with empty data
-	mStore.On("Set", store.ProposalRecoveryMetadata, []ocr2keepers.UpkeepPayload{}).Times(3)
+	mStore.On("Set", store.ProposalRecoveryMetadata, testData).Times(3)
```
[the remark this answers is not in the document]
shileiwill: Aha, I changed this! I think this is a side effect after I add the preprocessor. We can talk about this offline soon.

pkg/v3/plugin/ocr3.go
shileiwill: I can just use the PreProcessor interface.

pkg/v3/flows/recovery.go
```diff
@@ -0,0 +1,63 @@
...
+	"time"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/postprocessors"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/service"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v3/tickers"
+)
+
+func newRecoveryProposalFlow(preprocessors []ocr2keepersv3.PreProcessor[ocr2keepers.UpkeepPayload], rs ResultStore, ms MetadataStore, rp RecoverableProvider, recoveryInterval time.Duration, logger *log.Logger, configFuncs ...tickers.ScheduleTickerConfigFunc) (service.Recoverable, Retryer) {
```
EasterTheBunny, line 15: I just noticed that the recovery proposal flow doesn't use the result store anymore. We will need to listen for result TTL events, but that can be another interface added later. You can remove the result store dependency.

EasterTheBunny (changes requested):

EasterTheBunny (approved):

PR #188, opened 2023-07-19
use block history to find coordinated block

EasterTheBunny:


pkg/v3/plugin/coordinateblock.go
```diff
@@ -0,0 +1,92 @@
...
+		if err != nil {
+			continue
+		}
+
+		// TODO: don't use values at index
+		height := heightHash{v, values[1]}
+
+		if _, present := p.seen[height]; !present {
+			p.seen[height] = struct{}{}
+			p.recent[height]++
```
shileiwill, line 60: so we dont care about the value of `seen` map, it is just empty struct. And `recent` counts from 0

shileiwill (approved):

shileiwill (approved):

PR #189, opened 2023-07-20
create log upkeep state enum

FelixFan1992:


pkg/basetypes.go
```diff
@@ -154,3 +154,10 @@ type ReportedUpkeep struct {
 }
 
 type BlockHistory []BlockKey
+
+type LogUpkeepState uint8
+
+const (
+	Eligible LogUpkeepState = iota
+	Performed
+)
```
EasterTheBunny, line 163: This is for the upkeep state interface correct? I think the goal is to have only final states so let's leave 'Eligible' out.

EasterTheBunny (changes requested):

pkg/basetypes.go
FelixFan1992: done

pkg/basetypes.go
```diff
@@ -154,3 +154,25 @@ type ReportedUpkeep struct {
...
+	SelectByID(ID string) (*UpkeepPayload, UpkeepState, error)
+	// SelectByBlock retrieves upkeep states at a specific block
+	SelectByBlock(block int64) ([]*UpkeepPayload, []UpkeepState, error)
+	// SelectByBlockRange retrieves upkeep states within block range from start (inclusive) to end (exclusive)
+	SelectByBlockRange(start, end int64) ([]*UpkeepPayload, []UpkeepState, error)
+	// SelectByUpkeepID retrieves upkeep states for an upkeep
+	SelectByUpkeepID(upkeepId *big.Int) ([]*UpkeepPayload, []UpkeepState, error)
+	// SelectByUpkeepIDs retrieves upkeep states for provided upkeeps
+	SelectByUpkeepIDs([]*big.Int) ([]*UpkeepPayload, []UpkeepState, error)
+}
```
EasterTheBunny, line 173: I don't think we need this interface here. It isn't used by the plugin so should probably be put in the core node instead.

pkg/basetypes.go
```diff
@@ -154,3 +154,25 @@ type ReportedUpkeep struct {
...
+
+type UpkeepStateUpdater interface {
+	SetUpkeepState(UpkeepPayload, UpkeepState) error
+	SetUpkeepStates([]UpkeepPayload, []UpkeepState) error
+}
```
EasterTheBunny, line 178: The pattern for interfaces we are following is to put the interface in the same file as the function that requires it as a dependency. The main consumer of this interface will be in `pkg/v3/plugin/delegate.go` in the config.

Also, this interface should be reduced to only `SetUpkeepState` for now as that is the only function the plugin will need to start with. The implementation of this interface can still include the other if desired, but let's keep the interface consumed by the plugin simple for now.

EasterTheBunny (changes requested):

pkg/basetypes.go
FelixFan1992: done

pkg/basetypes.go
FelixFan1992: done

EasterTheBunny:
Finally, I think this needs to be rebased on top of main instead of develop. I will be deleting develop.

PR #191, opened 2023-07-20
Outcome History Ring Buffer

EasterTheBunny:
Block coordination happens independently of other data coordination, but is a required value for other processes. To correctly be able to merge coordinated blocks and coordinated recoverable ids, outcome history is required. This can either be local state or we can use the outcome to store previous coordinated values.

This commit introduces a ring buffer of outcome history that is maintained collectively by all nodes. This allows all nodes to coordinate on various bits of data as well as recent history.

### DIFF
--- a/pkg/v3/plugin/history_test.go
+++ b/pkg/v3/plugin/history_test.go
@@ -1,0 +1,160 @@
+package plugin
+
+import (
+	"testing"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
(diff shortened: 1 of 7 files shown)


pkg/v3/plugin/history_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"testing"
+
+	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg"
+	ocr2keepersv3 "github.com/smartcontractkit/ocr2keepers/pkg/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUpdateHistory(t *testing.T) {
```
shileiwill, line 11: thanks for the tests!

shileiwill (approved):

PR #194, opened 2023-07-21
[QA-1200] add sonarqube integration

anirudhwarrier:


.github/workflows/go.yml
```diff
@@ -19,10 +19,23 @@ jobs:
         go-version-file: "go.mod"
 
     - name: golangci-lint
-      uses: golangci/golangci-lint-action@v3
+      uses: smartcontractkit/golangci-lint-action@54ab6c5f11d66a92d14c3f7cc41ea13f676644bd
```
ferglor, line 22: What's the significance of this particular version?

.github/workflows/go.yml
```diff
@@ -19,10 +19,23 @@ jobs:
...
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-        version: latest
+        version: v1.52.1
+        allow-extra-out-format-args: true
+        args: --out-format checkstyle:golangci-lint-report.xml
+
+    - name: Print golangci lint report
+      if: always()
+      run: test -f golangci-lint-report.xml && cat golangci-lint-report.xml || true
```
ferglor, line 31: Is the `|| true` needed so that this step always succeeds, whether the report file exists or not?

.github/workflows/go.yml
```diff
@@ -66,7 +85,14 @@ jobs:
         go-version-file: "go.mod"
 
     - name: Run Race Tests
-      run: go test -race ./...
+      run: go test -race ./... -coverpkg=./... -coverprofile=race_coverage.txt
```
ferglor, line 88: Is it possible to run the race tests a number of times?

.github/workflows/go.yml
anirudhwarrier: This is the version used in the main chainlink repo as well. 
https://github.com/smartcontractkit/chainlink/blob/develop/.github/workflows/ci-core.yml#L82-L85
This change is to support multiple --out-format args

.github/workflows/go.yml
anirudhwarrier: Yes that is the idea. Again copied over from chainlink repo as it is there.

.github/workflows/go.yml
ferglor: It might be worth having something like 10 rounds for race tests; testing locally, I do this to make sure any race issues have been fixed, but 100 iterations might be overkill for CI 😅 

```bash
for i in {1..100}; do go clean -testcache; go test ./... -race; done
```

ferglor (approved):

PR #200, opened 2023-07-25
Plugin input validation

EasterTheBunny:


pkg/basetypes.go
```diff
@@ -194,6 +226,18 @@ type Trigger struct {
 	Extension interface{}
 }
 
+func ValidateTrigger(t Trigger) error {
+	if t.BlockNumber == 0 {
```
ferglor, line 230: Do we need to check against the indefinite blocking key also?

ferglor (approved):

PR #205, opened 2023-07-26
Finalize e2e docs

amirylm:


docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+Processes latest logs for active upkeeps. It does the following procedures:
+
+- Called every second by Log Trigger Ticker with latest logs as full payloads.
+- Filter out logs with block number that is older than threshold
+    - These are expected to be picked up by the log recoverer 
+- Pre-processes to filter the logs already present in coordinator
+- Calls runner with upkeep payload
+- If upkeep is eligible enqueue into result store
+- If runner gave an error, put back into log retry ticker, to be performed after a fixed timeout (can extend to exponential backoff)
```
EasterTheBunny, line 247: Exponential backoff isn't necessary here as the retry window is very small anyway. The retry cadence is to wait 10 seconds, try again, and repeat for up to 5 minutes. This is configurable.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+<aside>
+💡 Note: Duplicate logs can enter this flow, this component should be able to handle that. This component can maintain in-memory state for processed logs and can also use that state for exponential backoff
+</aside>
+
+### Retry Observer
+
+**TODO**
```
EasterTheBunny, line 255: The retry observer uses a schedule ticker as the driver with a coordinator preprocessor and runner. For each payload that enters the retry observer a wait time of 10 seconds is applied before being forwarded on to the runner. In postprocessing, if a result is retryable and the max retry time of 5 minutes has not passed, it is added back to the same retry observer to wait again. If the max retry time has elapsed, the result is passed to a recovery proposal observer. If the result is eligible, it is sent to the result store.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Recovery Observer
+
+This flow gets as input an upkeepID and a logIdentifier. It is a backup flow which verifies that a log was fully processed. Logs can be put here from the node itself (through Log Recoverer) or from other nodes (through Plugin)
```
EasterTheBunny, line 259: There are two parts to recovery:
- recovery proposal
- recovery with coordinated block

The first, `recovery proposal`, collects recoverable items from the node itself and runs a retry attempt cadence similar to the retry observer. The sources of recoverables in this component are
- recoverable provider
- result store TTL
- retry observer failures
- main log flow if retry observer rejects a result
- final recovery with coordinated block

No recovery proposals run through the pipeline during their retry cadence. Instead they are surfaced to the network and achieve quorum per recoverable. For each item added to this flow, a wait time is applied before attempting to put it into the metadata store. This wait time is once per minute with a maximum of 24 hours. Items can be added multiple times to the recovery proposal flow from multiple sources without altering wait times. This allows the recovery proposal to accept repeated proposals and provide a unified output.

The second, `recovery with coordinated block`, has an input of recoverable proposals that have achieved quorum, applies the latest coordinated block upon entry to the flow and are passed through the runner. If the results are eligible, they are forwarded to the result store. Otherwise, the results are sent back to the recovery proposal flow.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+- Input is an upkeepID, logTxHash, logIndex (optional checkBlockNum, checkBlockHash when passed from plugin)
+- Checks whether log should be processed
+    - Whether its part of the filters for that upkeep and it is still present on chain (DB call to log poller)
+    - Whether it is older from the latest block within a threshold (To prevent malicious node requesting recovery on a new log)
```
EasterTheBunny, line 264: This is not currently part of the recovery observer. These protections were added to the check pipeline as input validation where the pipeline won't return a result that is outside it operating parameters.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+- Checks whether the log has been already processed
+    - If node locally checked (logTxHash, logIndex) and got a non eligible result then skip (DB call)
+    - If chain has a perform for the log then skip (DB call which looks at upkeepPerformed logs)
```
EasterTheBunny, line 267: These are both TODO items

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+- If checkBlockNum not present: puts into an **in-memory recovery queue** which is then read in the plugin and put back in the flow with a coordinated checkBlock
+- If checkBlockNum present
+    - get payload from log provider for (upkeepID, logTxHash, logIndex)
+    - call runner with payload
+    - upon result, if eligible add into result store
+    - non-eligible are written to DB
+    - If runner gave an error then it is ignored (it might get picked up again by log recoverer, but we do not keep retrying here)
```
EasterTheBunny, line 274: The plugin does continue retrying by putting the retryable result back into the recovery proposal flow.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+<aside>
+💡 Note: Duplicate logs can enter this flow, this component should be able to handle that
+</aside>
```
EasterTheBunny, line 278: This is allowed within the recovery proposal observer / flow

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+#### Observation
+
+Observation starts with a processing of the previous outcome:
+- Remove agreed upon finalized results from result store
+- For `acceptedSamples` (bound to a trigger)
+    - Remove from metadata store
+    - Enqueue (trigger, upkeepID) into conditional observer
+- For `acceptedRecoveryLogs` (bound to a trigger)
+    - Remove from metadata store
+    - Enqueue (trigger, upkeepID) into recovery observer
```
EasterTheBunny, line 343: Both of these processes also use the latest coordinated block from the previous outcome. If there isn't a previous coordinated block in the recent outcome history, the results are ignored.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+Then we do the following for current observation:
+- Query results from result store giving seqNr as pseudoRandom seed and predefined limit. Filter results using coordinator and add them to observation
+- Query from metadata store for sample and recovery instructions using seqNr as pseudoRandom seed within limits, filter using coordinator and add them to observation
+- Query latest block number and hashes from local chain, add them to observation
+
+#### Outcome
+
+- Derive latest blockNumber and blockHash
```
EasterTheBunny, line 352: This is not done in every outcome. When all the nodes agree that block coordination should be done, then they do it using the block history. The block that achieves quorum is the most recent block/hash that the majority of nodes have in common.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+- Any result which has f+1 agreement is added to finalized result
+- All samples are collected from observations within limits, deduped and filtered from existing `acceptedSamples`. These are then added to `acceptedSamples` in the outcome bound to the current latestBlockNumber and hash.
```
EasterTheBunny, line 354: filtering is not done at this point.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+    - `acceptedSamples` is a ring buffer where samples are held for ‘x’ (~30) rounds so that they do not get bound to a new blockNumber for some time
```
EasterTheBunny, line 355: `acceptedSamples` are not currently a ring buffer by themselves. history in general is a ring buffer to allow access to the latest coordinated block. at the moment, this history is very heavy handed and there should be a TODO to reduce the data footprint of it.

EasterTheBunny (commented):
I left a lot of comments and I'm tired now. They should help clear up any small remaining details.

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Reliability Guarantees
+
+**Log triggers** 
+Out of n=10 nodes every node listens to configured user log, as soon f+1=3 nodes see the log and agree on checkPipeline, it will be performed on chain. We can handle up to 7 nodes missing a log and handle capacity of upto 10 log trigger upkeeps with rate limit per upkeep of 5 logs per second.
+
+**Conditionals** 
+Every upkeep’s condition will be checked at least once by the network every ~3 seconds, handling up to f+1=3 nodes being down. Once condition is eligible, every node evaluates the checkPipeline, as soon as f+1=3 nodes agree, it will be performed on chain. We can handle handle capacity of upto 500 conditional upkeeps.
+    
+The protocol will be functional as long as > 6 ((n+f)/2) nodes are alive and participating within the p2p network (required for ocr3 consensus).
```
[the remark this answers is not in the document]
infiloop2: Do you have any reference link to research terminology? 

I had heard slightly different meanings of consensus and agreement in past

Regardless, in this particular line, we do need n+f/2 nodes to agree on an outcome so I'll leave consensus here.
In other places I'll replace by agreement of quorum of nodes

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+Returns list of results for each payload. Result can be **eligible** with upkeepResult / **non-eligible** with failure reason / **error.** Error can be retryable if it’s transient or non retryable in case the checkBlockNum is too old.
+
+### Runner
+
+This component is responsible for parallelizing upkeep executions and providing a single interface to different components maintaining a shared cache
+
+- Takes a list of upkeepPayloads, calls CheckPipeline asynchronously with upkeeps being batched in a single pipeline execution
+- Maintains in-memory cache of non-errored pipeline executions indexed on (trigger, upkeepID). Directly uses that result instead of a new execution if available
+- Allows for repeated calls for the same (trigger, upkeepID). If an execution is in progress then it doesn’t start a new execution
```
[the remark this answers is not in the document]
infiloop2: @amirylm let's create a separate list of known TODOs in code

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+- Does a callback when results are available with the result
+
+<aside>
+💡 Note: This component can be made synchronous instead of giving a callback while callers can call it asynchronously
+</aside>
+
+### Coordinator
+
+This component stores in-flight & confirmed reports in memory, and allows other components in the system to query that information. 
+It maintains a single item per `upkeepTriggerID`, and stores `confirmed` flag, the `upkeepID` and `trigger`.
```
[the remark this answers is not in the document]
infiloop2: It might make sense to split this into two separate components here

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+
+### Runner
+
+This component is responsible for parallelizing upkeep executions and providing a single interface to different components maintaining a shared cache
+
+- Takes a list of upkeepPayloads, calls CheckPipeline asynchronously with upkeeps being batched in a single pipeline execution
+- Maintains in-memory cache of non-errored pipeline executions indexed on (trigger, upkeepID). Directly uses that result instead of a new execution if available
+- Allows for repeated calls for the same (trigger, upkeepID). If an execution is in progress then it doesn’t start a new execution
+- Execution automatically fails after configured timeout. (~10s)
+- Does a callback when results are available with the result
```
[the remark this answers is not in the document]
infiloop2: Needs to be updated according to latest implementation

docs/PROTOCOL_v21.md
infiloop2: How does this work when block goes out of range? (e.g. on arbitrum we can only retry for 128 blocks ~60 seconds)

docs/PROTOCOL_v21.md
infiloop2: Needs to be thought through the implications of not doing filtering here

docs/PROTOCOL_v21.md
```diff
@@ -0,0 +1,375 @@
...
+Returns list of results for each payload. Result can be **eligible** with upkeepResult / **non-eligible** with failure reason / **error.** Error can be retryable if it’s transient or non retryable in case the checkBlockNum is too old.
+
+### Runner
+
+This component is responsible for parallelizing upkeep executions and providing a single interface to different components maintaining a shared cache
+
+- Takes a list of upkeepPayloads, calls CheckPipeline asynchronously with upkeeps being batched in a single pipeline execution
+- Maintains in-memory cache of non-errored pipeline executions indexed on (trigger, upkeepID). Directly uses that result instead of a new execution if available
+- Allows for repeated calls for the same (trigger, upkeepID). If an execution is in progress then it doesn’t start a new execution
+- Execution automatically fails after configured timeout. (~10s)
```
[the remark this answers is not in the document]
amirylm: The runner is the component that actually applies this timeout

docs/PROTOCOL_v21.md
infiloop2: I'm not sure why we started like that, we should update this to store the history that we require

docs/PROTOCOL_v21.md
infiloop2: Not fully understand why we need wait time before putting into metadata store

docs/PROTOCOL_v21.md
infiloop2: moved to log provider

docs/PROTOCOL_v21.md
infiloop2: Ended the flow here. The log will be picked up again by log recoverer

docs/PROTOCOL_v21.md
infiloop2: The sources of recoverables can also be simplified by only having the log recoverer responsible for all recovery by keeping a strict range. Will discuss with @amirylm

infiloop2 (approved):

infiloop2 (approved):

PR #207, opened 2023-07-26
Wire up samples and recovery

EasterTheBunny:
This commit wires together the metadata store and OCR for conditional flows and
recovery flows. When samples are detected in an outcome, a coordinated block is
searched for in the outcome history. If one is found, the samples are merged
with the latest coordinated block and pushed to the runner.

The same process exists for recovery items. When recovery items are detected in
an outcome, a coordinated block is searched for in the outcome history. If one
is found, the recovery items are merged with the latest coordinated block and
pushed to the runner.


pkg/v3/hooks/build/add_from_recovery.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	proposals := make([]ocr2keepers.CoordinatedProposal, 0)
+	for _, key := range cache.Keys() {
+		v, ok := cache.Get(key)
+		if !ok {
+			cache.Delete(key)
+		}
+
+		proposals = append(proposals, v)
```
ferglor, line 30: Should this be wrapped in an else? Do we want to append `v` even if the `Get` wasn't okay?

pkg/v3/ocrtypes.go
```diff
@@ -150,3 +157,93 @@ func DecodeAutomationOutcome(data []byte) (AutomationOutcome, error) {
 		NextIdx:      rawOutcome.NextIdx,
 	}, nil
 }
+
+func RecoveryProposalsFromOutcome(outcome AutomationOutcome) ([]ocr2keepers.CoordinatedProposal, error) {
+	var ok bool
+	var rawProposals interface{}
+	var proposals []ocr2keepers.CoordinatedProposal
```
ferglor, line 164: 🏄 These could be in a var block?

pkg/v3/ocrtypes.go
```diff
@@ -150,3 +157,93 @@ func DecodeAutomationOutcome(data []byte) (AutomationOutcome, error) {
...
+		return nil, fmt.Errorf("%w: coordinated proposals are not of type `CoordinatedProposal`", ErrWrongDataType)
+	}
+
+	return proposals, nil
+}
+
+func UpkeepIdentifiersFromOutcome(outcome AutomationOutcome) ([]ocr2keepers.UpkeepIdentifier, error) {
+	var ok bool
+	var rawProposals interface{}
+	var proposals []ocr2keepers.UpkeepIdentifier
```
ferglor, line 182: Same here

pkg/v3/flows/conditional.go
```diff
@@ -57,56 +60,33 @@ func NewConditionalEligibility(
...
+		mStore:  ms,
 		builder: builder,
 		final:   point,
 		logger:  logger,
 	}, []service.Recoverable{svc0, svc1}, err
 }
 
 func (flow *ConditionalEligibility) ProcessOutcome(outcome ocr2keepersv3.AutomationOutcome) error {
-	var ok bool
+	samples, err := ocr2keepersv3.UpkeepIdentifiersFromOutcome(outcome)
```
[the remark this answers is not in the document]
EasterTheBunny: I put it where it is because I thought the scope of the function was a general helper function for accessing data in an outcome. While `flows` is the only package that currently uses it, I would prefer it be available for all because the scope isn't special for `flows`.

pkg/v3/hooks/build/add_from_recovery.go
EasterTheBunny: You're right. I'll fix this.

pkg/v3/flows/conditional.go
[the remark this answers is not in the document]
EasterTheBunny: Maybe it should be a static function of `AutomationOutcome`?

ferglor (approved):

PR #208, opened 2023-07-27
Integrate the conditional coordinator for v3

ferglor:
In this PR, we're lifting the conditional coordinator from v2, and reworking it to fit in with the v3 coordinator interface

The key differences are:

- UpkeepKeys are now replaced with UpkeepPayloads
- The stale and perform logs are now provided from an events provider instead of a logs provider


pkg/v3/coordinator/conditional_report_coordinator.go
```diff
@@ -0,0 +1,342 @@
...
+	return nil
+}
+
+// IsTransmissionConfirmed returns whether the upkeep was successfully
+// completed or not
+func (rc *conditionalReportCoordinator) IsTransmissionConfirmed(key ocr2keepers.UpkeepPayload) bool {
+	// key is confirmed if it both exists and has been confirmed by the log
+	// poller
+	confirmed, ok := rc.activeKeys.Get(string(key.Upkeep.ID))
+	return !ok || (ok && confirmed)
```
ferglor, line 141: Is this existing functionality correct - if the upkeep ID doesn't exist in `activeKeys` we consider it completed?

EasterTheBunny: Yes. The active keys has a timeout, but once the key is gone we assume that it is completed. This could happen if a transaction is stuck in the mempool for a long period of time. We then assume completion and allow further checks and performs of the upkeep. If the stuck transaction does get transmitted after a more up-to-date transaction does, the old one will result in a stale event. This results in a bit of gas waste in the scenario, but is functional.

PR #211, opened 2023-07-28
Connect recovery flow components

EasterTheBunny:
Items in the final recovery flow with coordinated block can still fail and end up recoverable again. By connecting the final recovery with the recovery proposal flow, recoverable items can be re-introduced to the recovery proposal process again and not need to rely as much on the recoverable provider.

### DIFF
--- a/pkg/v3/postprocessors/retry.go
+++ b/pkg/v3/postprocessors/retry.go
@@ -30,19 +30,23 @@ func (p *retryPostProcessor) PostProcess(_ context.Context, results []ocr2keeper
 }
 
 func (p *retryPostProcessor) attemptRetry(res ocr2keepers.CheckResult) error {
-	err := p.retryer.Retry(res)
-	if err == nil {
-		return nil
+	var err error
+
+	if p.retryer != nil {
+		err = p.retryer.Retry(res)
+		if err == nil {
+			return nil
+		}
 	}
 
-	if errors.Is(err, tickers.ErrSendDurationExceeded) {
+	if err == nil || errors.Is(err, tickers.ErrSendDurationExceeded) {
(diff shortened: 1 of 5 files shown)


ogtownsend (approved):
lgtm, but might be good to have someone more qualified check it out :P

pkg/v3/postprocessors/retry.go
```diff
@@ -30,19 +30,23 @@ func (p *retryPostProcessor) PostProcess(_ context.Context, results []ocr2keeper
 }
 
 func (p *retryPostProcessor) attemptRetry(res ocr2keepers.CheckResult) error {
-	err := p.retryer.Retry(res)
-	if err == nil {
-		return nil
+	var err error
+
+	if p.retryer != nil {
```
ferglor, line 35: How likely is it that the retry post processor will have a nil retryer?

pkg/v3/postprocessors/retry.go
```diff
@@ -30,19 +30,23 @@ func (p *retryPostProcessor) PostProcess(_ context.Context, results []ocr2keeper
...
+		err = p.retryer.Retry(res)
+		if err == nil {
+			return nil
+		}
 	}
 
-	if errors.Is(err, tickers.ErrSendDurationExceeded) {
-		if err := p.recoverer.Retry(res); err != nil {
+	if err == nil || errors.Is(err, tickers.ErrSendDurationExceeded) {
```
ferglor, line 42: I find this logic pretty tricky to follow; can we have full test coverage for this function so that we have all possible flows covered? Or could we rewrite this so that the error cases are more closely tied to the calls that return those errors?

PR #220, opened 2023-08-02
Gaps from design

infiloop2:


pkg/basetypes.go
```diff
@@ -232,6 +232,8 @@ type Trigger struct {
 	BlockNumber int64
 	// BlockHash is the block hash of the corresponding block
 	BlockHash string
+	// TODO: Improve this and other types to be more concrete so it can be easily decoded/encoded into bytes
```
infiloop2, line 235: Keeping as is for now.

pkg/v3/coordinator/conditional_report_coordinator.go
```diff
@@ -192,6 +193,7 @@ func (rc *conditionalReportCoordinator) checkEvents(ctx context.Context) error {
 		}
 
 		if confirmed, ok := rc.activeKeys.Get(string(evt.UpkeepID)); ok {
+			// TODO: This all logic can be removed
```
infiloop2, line 196: Can do as a code clean up as low priority

PR #222, opened 2023-08-02
Add a mercury lookup preprocessor

ferglor:
In this PR, we're adding a new preprocessor in the plugin, which will be used in the conditional eligibility flow, that will enable mercury lookup on the upkeep payloads, if mercury lookup is enabled on the offchain config

### DIFF
--- a/pkg/basetypes.go
+++ b/pkg/basetypes.go
@@ -192,6 +192,8 @@ type UpkeepPayload struct {
 	CheckData []byte
 	// Trigger is the event that triggered the upkeep to be checked
 	Trigger Trigger
+	// MercuryLookup a flag that determines if a mercury lookup should be performed
+	MercuryLookup bool
 }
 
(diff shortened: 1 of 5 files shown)


pkg/basetypes.go
```diff
@@ -192,6 +192,8 @@ type UpkeepPayload struct {
 	CheckData []byte
 	// Trigger is the event that triggered the upkeep to be checked
 	Trigger Trigger
+	// MercuryLookup a flag that determines if a mercury lookup should be performed
+	MercuryLookup bool
```
ferglor, line 196: How we communicate `MercuryLookup` on the upkeep payload was a little up in the air; so I opted for the most straightforward approach here, just to get something in place. We can discuss if we want a map to store config values, and where exactly within the payload that map should live if neede 👍

EasterTheBunny: Looks like it will work just fine.

EasterTheBunny (approved):
LGTM

PR #227, opened 2023-08-06
Add workID and config for passing functions from core to ocr2keepers

infiloop2:
- Add workID in types
- Add upkeepTypeGetter to be passed from core, use this within log coordinator
- Remove checkBlock from upkeepPayload

Corresponding CL repo PR https://github.com/smartcontractkit/chainlink/pull/10095


pkg/basetypes.go
```diff
@@ -1,17 +1,20 @@
...
-	"github.com/ethereum/go-ethereum/crypto"
 )
 
 type UpkeepIdentifier []byte
 
+type UpkeepType uint8
+
+const (
+	ConditionTrigger UpkeepType = iota
+	LogTrigger
```
ferglor, line 15: Just thinking about this; won't the `0` value assigned to `ConditionTrigger` mean that invalid/incomplete `UpkeepIdentifier` instances on upkeeps will be treated as `ConditionTrigger` types? 🤔 Like should we avoid relying on "default" values to determine validity? Not sure if that's something we need to consider/care about when testing error scenarios but just wanted to put it out there!

ferglor (approved):

pkg/basetypes.go
amirylm: Good point

amirylm (approved):

pkg/basetypes.go
infiloop2: good point to think through. Upkeep identifiers are uint256 and as long as they are in the range of 0 to 2^256-1 they should be valid.

With new trigger types we defined an encoding where 4th to 14th bytes should be 0 for the upkeepID and 15th byte should contain the uint8 type. However we have old upkeepIDs defined in the system which might not follow the above pattern, there we want default to be conditionals.

Logic on contract here https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/dev/automation/2_1/KeeperRegistryBase2_1.sol#L683-L694 

I haven't verified if the offchain logic matches this, will create a ticket for it

PR #232, opened 2023-08-09
Finalize plugin - v3

ferglor:


pkg/v3/plugin/ocr3.go
```diff
@@ -276,24 +218,20 @@ func (plugin *ocr3Plugin) ShouldTransmitAcceptedReport(_ context.Context, seqNr
...
-		allConfirmationsFalse := true
-		for _, coord := range plugin.Coordinators {
-			if coord.IsTransmissionConfirmed(upkeep) {
-				allConfirmationsFalse = false
-			}
-
-			plugin.Logger.Printf("checking transmit of upkeep '%s' %t", upkeep.UpkeepID, coord.IsTransmissionConfirmed(upkeep))
+		shouldTransmit := plugin.Coordinator.ShouldTransmit(upkeep)
+		if !shouldTransmit {
+			transmit = false
```
ferglor, line 227: The `allConfirmationsFalse` flag in the previous version of the code confused me here; is this new functionality correct? I.e. if even one upkeep receives a `false` response from `ShouldTransmit`, we should return false from this function?

pkg/v3/store/metadata_store.go
```diff
@@ -0,0 +1,162 @@
...
+	RemoveProposalConditional(...types.CoordinatedProposal) []types.CoordinatedProposal
+	Start(context.Context) error
+	Close() error
+}
+
+type metadataStore struct {
+	blocks                     *tickers.BlockTicker
+	blockHistory               types.BlockHistory
+	blockHistoryMutex          sync.RWMutex
+	proposalLogRecovery        *util.Cache[types.CoordinatedProposal]
```
ferglor, line 41: This was pretty surprising; the value stored for `ProposalRecoveryMetadata` wasn't a list of objects, but a separate cache? I also couldn't find a cache cleaner for this cache, so I'm not sure if that could have potentially caused a memory leak? 

Beyond retrieving the cache, there didn't seem to be use of a mutex when reading/updating values, so I'm not sure if that's something we need to consider here or not?

amirylm (approved):

PR #233, opened 2023-08-09
Split out v2 packages

infiloop2:
- Move packages in /pkg to /v2 to have a clear split between v2 and v3
- only utils package is shared

- config was shared between packages, copy it into both v2 and v3. Ensure files in v2/folder refer to v2 config, and in v3 folder refer to v3 config

### DIFF
--- a/cmd/simv3/main.go
+++ b/cmd/simv3/main.go
@@ -20,7 +20,6 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/run"
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/simulators"
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/telemetry"
-	"github.com/smartcontractkit/ocr2keepers/pkg/v2/encoding"
 )
 
 var (
@@ -135,5 +134,4 @@ func main() {
 
 type fullEncoder struct {
 	simulators.SimulatedReportEncoder
-	encoding.BasicEncoder
 }
--- a/cmd/simv3/simulators/registry_test.go
+++ b/cmd/simv3/simulators/registry_test.go
@@ -9,7 +9,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 
-	"github.com/smartcontractkit/ocr2keepers/pkg/v2/encoding"
 	ocr2keepers "github.com/smartcontractkit/ocr2keepers/pkg/v3/types"
 )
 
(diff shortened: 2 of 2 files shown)


cmd/simv3/main.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/run"
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/simulators"
 	"github.com/smartcontractkit/ocr2keepers/cmd/simv3/telemetry"
-	"github.com/smartcontractkit/ocr2keepers/pkg/encoding"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v2/encoding"
```
infiloop2, line 23: not sure why v2 encoding was referenced in simv3, ignoring it for now

cmd/simv3/simulators/registry_test.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 
-	"github.com/smartcontractkit/ocr2keepers/pkg/encoding"
+	"github.com/smartcontractkit/ocr2keepers/pkg/v2/encoding"
```
infiloop2, line 12: same. not sure why v2 encoding was referenced in simv3, ignoring it for now

pkg/v3/coordinator/conditional_report_coordinator.go
```diff
@@ -42,7 +42,7 @@ var (
 	cadence               = time.Second
 )
 
-//go:generate mockery --name Encoder --srcpkg "github.com/smartcontractkit/ocr2keepers/pkg/coordinator" --case underscore --filename encoder.generated.go
+//go:generate mockery --name Encoder --srcpkg "github.com/smartcontractkit/ocr2keepers/pkg/v3/coordinator" --case underscore --filename encoder.generated.go
```
infiloop2, line 45: this was incorrect before

cmd/simv3/main.go
ferglor: The few v2 components in the v3 sim that I saw as part of the type refactoring, weren't actually being used, so I guess it was maybe just a copy-paste artifact?

ferglor (dismissed):

cmd/simv3/main.go
infiloop2: i tried removing them but ran into errors so let them be for now

amirylm (approved):

cmd/simv3/main.go
infiloop2: actually was able to remove them now i think :D

ferglor (approved):

PR #236, opened 2023-08-15
remove generic result type on observer

RyanRHall:


pkg/v3/observer.go
```diff
@@ -28,12 +28,12 @@ type Runner interface {
 	CheckUpkeeps(context.Context, ...ocr2keepers.UpkeepPayload) ([]ocr2keepers.CheckResult, error)
 }
 
-type Observer[T any, R any] struct {
+type Observer[T any] struct {
```
infiloop2, line 31: if you want you can also cleanup the input type which is always an UpkeepPayload

infiloop2 (approved):
looks good. Added suggestion for further cleanup of generic which would make this cleaner

PR #238, opened 2023-08-15
Async ticker observe

infiloop2:
Within ticker, call observer.process asynchronously since it can take a while and end up blocking further ticks


pkg/v3/tickers/time.go
```diff
@@ -62,10 +62,13 @@ func (t *timeTicker[T]) Start(pctx context.Context) error {
...
-			if err := t.observer.Process(ctx, tick); err != nil {
-				t.logger.Printf("error processing observer: %s", err.Error())
-			}
+			// observer.Process can be a heavy call taking upto ObservationProcessLimit seconds
+			// so it is run in a separate goroutine to not block further ticks
+			go func(o observer[T], logger *log.Logger) {
+				if err := o.Process(ctx, tick); err != nil {
+					logger.Printf("error processing observer: %s", err.Error())
+				}
+			}(t.observer, t.logger)
```
infiloop2: two things to discuss here:
1. Should we limit the number of goroutines being fired here
2. Should we wait for these gorountines to finish upon close?

Also we can potentially move this to Observer.Process so that the timeout is more explicit for this routine

amirylm (dismissed):

pkg/v3/tickers/time.go
infiloop2: Discussed offline. Added exploratory task to limit the number of goroutines

amirylm (approved):

PR #239, opened 2023-08-15
make CheckResult print fields in human-readable format

RyanRHall:
adds `String()` methods to the following structs:
* `CheckResult`
* `Trigger`
* `LogTriggerExtension`

### DIFF
--- a/pkg/v3/types/basetypes.go
+++ b/pkg/v3/types/basetypes.go
@@ -1,9 +1,31 @@
 package types
 
 import (
+	"encoding/hex"
 	"fmt"
 	"math/big"
+	"strings"
 )
+
+var checkResultStringTemplate = `{
+	"PipelineExecutionState":%d,
+	"Retryable":%v,
+	"Eligible":%v,
+	"IneligibilityReason":%d,
+	"UpkeepID":%s,
+	"Trigger":%s,
+	"WorkID":"%s",
+	"GasAllocated":%d,
+	"PerformData":"%s",
+	"FastGasWei":%s,
+	"LinkNative":%s
+}`
+
+func init() {
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, " ", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\t", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\n", "", -1)
+}
 
 type UpkeepType uint8
 
@@ -37,6 +59,7 @@ const (
 // UpkeepIdentifier is a unique identifier for the upkeep, represented as uint256 in the contract.
 type UpkeepIdentifier [32]byte
 
+// String returns a base 10 numerical string representation of the upkeep identifier.
 func (u UpkeepIdentifier) String() string {
 	return u.BigInt().String()
 }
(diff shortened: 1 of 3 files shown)


pkg/v3/types/basetypes.go
```diff
@@ -1,8 +1,11 @@
 package types
 
 import (
+	"encoding/json"
 	"fmt"
 	"math/big"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
```
infiloop2, line 8: We don't want to import chain specific packages in this repo here unfortunately

pkg/v3/types/basetypes.go
```diff
@@ -150,6 +154,36 @@ func (r CheckResult) UniqueID() string {
...
+		PerformData string // 0x-prefixed hex string
+		FastGasWei  string // base 10 numerical string
+		LinkNative  string // base 10 numerical string
+	}{
+		CheckResultAlias: (CheckResultAlias)(r),
+		UpkeepID:         r.UpkeepID.String(),
+		PerformData:      hexutil.Encode(r.PerformData),
+		FastGasWei:       r.FastGasWei.String(),
+		LinkNative:       r.LinkNative.String(),
+	}
```
infiloop2, line 174: not sure if we should go such a heavy way to create string, we can simply derive this from fields instead of creating a new struct and json marshalling it which is expensive. See uniqueID() function

pkg/v3/types/basetypes.go
ferglor: I wonder is there a way we can add a deny list for packages we want to prohibit in keepers 🤔

RyanRHall: that was my understanding, but then i noticed we're still using go-ethersum in several places in the codebase (although it looks like these are all under "sim" - which I assume is only used in testing?

will remove this instance!

pkg/v3/types/basetypes.go
RyanRHall: fair, will rework

pkg/v3/types/basetypes.go
infiloop2: yep, sim is just for testing

pkg/v3/types/basetypes.go
```diff
@@ -150,6 +173,14 @@ func (r CheckResult) UniqueID() string {
 	return fmt.Sprintf("%x", resultBytes)
 }
 
+func (r CheckResult) String() string {
+	return fmt.Sprintf(
+		checkResultStringTemplate, r.PipelineExecutionState, r.Retryable, r.Eligible,
+		r.IneligibilityReason, r.UpkeepID, r.Trigger, r.WorkID, r.GasAllocated,
```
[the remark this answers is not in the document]
infiloop2: or maybe it works out of the box as I see it works in tests

pkg/v3/types/basetypes.go
```diff
@@ -1,10 +1,32 @@
...
+	"PerformData":"%s",
+	"FastGasWei":%s,
+	"LinkNative":%s
+}`
+
+func init() {
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, " ", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\t", "", -1)
+	checkResultStringTemplate = strings.Replace(checkResultStringTemplate, "\n", "", -1)
+}
```
infiloop2, line 28: not sure if this is a common pattern we don't use it in other packages. I don't have much experience here though will defer to @ferglor or @amirylm

pkg/v3/types/basetypes.go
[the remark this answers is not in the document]
RyanRHall: yeah because we're dong "%s"

pkg/v3/types/basetypes.go
RyanRHall: oh wow you're right. I think this is a pretty common go-ism, though I can remove it if y'all prefer. fwiw it seems we use init functions all over core.

ferglor: Yeah I think using init is fine; I take it the idea here is to ensure the templated string is collapsed into a single line for printing?

ferglor (approved):
It's so frustrating that we can't use something more convenient and expensive like json encoding but this is cool

pkg/v3/types/basetypes.go
RyanRHall: @ferglor yeah that was the idea

PR #240, opened 2023-08-16
Remove the block ticker

ferglor:
In this PR, we are removing the block ticker, and hooking the block subscriber directly into the metadata store

We had reports of high CPU usage in our staging environment, and looking at the CPU profile, we saw this:

image

In the block ticker's `Start` function, we have an infinite for loop with a select statement. The select statement has a non-empty default branch, which itself has an empty default branch. In the right conditions, this is equivalent to running `for {}`, until one of the `select` cases activates.

The block ticker's role was to read from the block subscriber, and send the received values onto its own internal channel (which the metadata store read from). The block subscriber also had functionality that would buffer a single value in the event that the downstream channel could not be written to, which would be sent downstream when the upstream channel could not be read from.

Rather than working with this functionality to find a fix, we decided it made sense to just hook the block subscriber directly into the metadata store, and subsequently get rid of the block ticker entirely.


amirylm (approved):

infiloop2:
Can we verify once the integration tests on core pass with this change

infiloop2 (approved):

PR #241, opened 2023-08-16
Filter empty payloads from payload builder

infiloop2:


pkg/v3/types/basetypes.go
```diff
@@ -172,6 +172,11 @@ type UpkeepPayload struct {
 	CheckData []byte
 }
 
+// Determines whether the payload is empty, used within filtering
+func (p UpkeepPayload) IsEmpty() bool {
```
ferglor, line 176: Should we add a TODO here for comparing more fields in future if needed?

ferglor (approved):

pkg/v3/types/basetypes.go
infiloop2: I can add the fields right now but wasn't sure they were needed since wid should be the unique identifier for this struct. wdyt?

infiloop2: discussed offline, keeping just workID for now

PR #244, opened 2023-08-18
Clean up config pkg

infiloop2:
Removing unused configs for now. We can add them or new ones later as we see need


ferglor (approved):
The best type of PR 🙏

PR #251, opened 2023-08-23
Add the ineligible post processor to the log recovery proposal flow

ferglor:
In this PR, we're adding the ineligible post processor to the log recovery proposal flow

We're also fixing some flaky tests (race conditions), and skipping two recoverable service tests that can occasionally become deadlocked


pkg/v3/postprocessors/ineligible_test.go
```diff
@@ -0,0 +1,114 @@
...
+					Eligible:               true,
+				},
+				{
+					PipelineExecutionState: 0,
+					Eligible:               false,
+				},
+			},
+			expectsErr: true,
+			wantErr: errors.New(`upkeep state boom
+upkeep state boom`),
```
infiloop2, line 86: 💥

pkg/v3/postprocessors/ineligible_test.go
```diff
@@ -0,0 +1,114 @@
...
+					return nil
+				},
+			},
+			results: []types.CheckResult{
+				{
+					PipelineExecutionState: 0,
+					Eligible:               false,
+				},
+			},
+			wantLog: "post-processing 1 results, 0 ineligible",
```
infiloop2, line 38: should the log here indicate 1 ineligible?

infiloop2: ah i think there's a bug in ineligible processor when counting ineligible count

infiloop2 (changes requested):
Requesting changes to investigate ineligible processor log

pkg/v3/postprocessors/ineligible_test.go
ferglor: This is a multi error check, so the new line is needed

infiloop2: i was just pointing out the boom :p

pkg/v3/postprocessors/ineligible_test.go
ferglor: Should the ineligible loop then become:

```if res.PipelineExecutionState == 0 && !res.Eligible {
	err := p.stateUpdater.SetUpkeepState(ctx, res, ocr2keepers.Ineligible)
	if err != nil {
		merr = errors.Join(merr, err)
		continue
	}
	ineligible++
}```

infiloop2: yup i think so!

PR #258, opened 2023-08-31
Add a delimiter to the check result unique ID

ferglor:
In this PR, we're updating the `UniqueID()` function of the `CheckResult` type, to append a tab character `0x09` to the unique ID, after every append operation


pkg/v3/types/basetypes.go
```diff
@@ -152,24 +156,54 @@ type CheckResult struct {
...
+
 	resultBytes = append(resultBytes, big.NewInt(int64(r.Trigger.BlockNumber)).Bytes()...)
+	resultBytes = append(resultBytes, checkResultDelimiter)
+
 	if r.Trigger.LogTriggerExtension != nil {
 		// Note: We encode the whole trigger extension so the behaiour of
 		// LogTriggerExtentsion.BlockNumber and LogTriggerExtentsion.BlockHash should be
 		// consistent across nodes when sending observations
 		resultBytes = append(resultBytes, []byte(fmt.Sprintf("%+v", r.Trigger.LogTriggerExtension))...)
+		resultBytes = append(resultBytes, checkResultDelimiter)
```
infiloop2, line 186: can we append the delimiter outside the if?

infiloop2 (changes requested):
requesting changes for adding delimiters for nil fields

infiloop2 (approved):

PR #261, opened 2023-09-02
Add more information to debug logs

infiloop2:


pkg/v3/coordinator/coordinator.go
```diff
@@ -178,7 +178,7 @@ func (c *coordinator) checkEvents(ctx context.Context) error {
 					transmitBlockNumber:   event.TransmitBlock,
 				}, util.DefaultCacheExpiration)
 			} else {
-				c.logger.Printf("Got event in transaction %s from newer report (block %v) while waiting for (block %v)", hex.EncodeToString(event.TransactionHash[:]), event.CheckBlock, v.checkBlockNumber)
+				c.logger.Printf("Got event in transaction %s of type %d for workID %s from newer report (block %v) while waiting for (block %v)", hex.EncodeToString(event.TransactionHash[:]), event.Type, event.WorkID, event.CheckBlock, v.checkBlockNumber)
```
ferglor, line 181: 🏄‍♂️ instead of an int representing event type, would it be more useful to have a text based name for this value?

ferglor (approved):

pkg/v3/coordinator/coordinator.go
infiloop2: yup, right now i just copied this from line 173, can improve this further in followup

PR #262, opened 2023-09-04
Include log block hash in wid and make it a required field

infiloop2:


pkg/v3/types/trigger.go
```diff
@@ -76,6 +74,7 @@ func (e LogTriggerExtension) LogIdentifier() []byte {
 	binary.BigEndian.PutUint32(indexBytes, e.Index)
 	return bytes.Join([][]byte{
 		e.TxHash[:],
+		e.BlockHash[:],
```
ferglor, line 77: Is there any significance to the order here? Does the block hash need to go before the index bytes or can it be added at any point?

ferglor (dismissed):

pkg/v3/types/trigger.go
infiloop2: good point. this needs to match the dedupe ID on contract which is under discussion

infiloop2: aligned with contract dedupe key order in https://github.com/smartcontractkit/chainlink/pull/10476

RyanRHall (approved):

PR #264, opened 2023-09-05
Add overall limit on sampled upkeeps

infiloop2:
in https://github.com/smartcontractkit/chainlink/pull/10485 we are putting overall limits on logs surfaced on every tick. Also putting a limit on conditionals samples done in a tick. We already have a random order defined on upkeeps, just put a limit on the total upkeeps we select

If number of required samples go above this, we should ideally log an error (not able to do right now with current logger). But the behaviour would be that the upkeeps would be bit delayed in getting sampled, rather than bombarding runner with a lot of checks which would affect log triggers too

### DIFF
--- a/pkg/v3/flows/conditional.go
+++ b/pkg/v3/flows/conditional.go
@@ -19,6 +19,8 @@ import (
 const (
 	// This is the ticker interval for sampling conditional flow
 	SamplingConditionInterval = 3 * time.Second
+	// Maximum number of upkeeps to be sampled in every round
+	MaxSampledConditionals = 300
 	// This is the ticker interval for final conditional flow
 	FinalConditionalInterval = 1 * time.Second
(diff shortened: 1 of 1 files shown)


pkg/v3/flows/conditional.go
```diff
@@ -19,6 +19,8 @@ import (
 const (
 	// This is the ticker interval for sampling conditional flow
 	SamplingConditionInterval = 3 * time.Second
+	// Maximum number of upkeeps to be sampled in every round
+	MaxSampledConditionals = 300
```
ferglor, line 23: If this isn't used outside this package I would lower case it so that it's not exported

ferglor (approved):

pkg/v3/flows/conditional.go
infiloop2: agree, but currently i would prefer to keep it consistent with other consts in package (which are also used just within the package but capitalised)

PR #269, opened 2023-09-11
Coordinator: ignore visited events

amirylm:


pkg/util/cache_clean.go
```diff
@@ -0,0 +1,46 @@
+package util
+
+import "time"
+
+// CacheCleaner is an interface for cleaning a cache
+type CacheCleaner interface {
```
ferglor, line 6: Is there a reason why we need the `CacheCleaner` as well as the `IntervalCacheCleaner`? Are we going to remove the interval cache cleaner when the cleanable gets removed as part of the v2 cleanup?

pkg/v3/coordinator/coordinator.go
```diff
@@ -234,7 +249,8 @@ func (c *coordinator) Start(pctx context.Context) error {
 		return fmt.Errorf("process already running")
 	}
 
-	go c.cacheCleaner.Run(c.cache)
+	go c.cacheCleaner.Start(c.cache)
+	go c.visitedCleaner.Start(c.visited)
```
ferglor, line 253: Would it make sense to have the cleaner as part of the cache itself? It's easy to forget to run a cache cleaner with this approach and cause memory leaks, and as far as I'm aware, the majority, if not all caches that we use, have an associated cleaner that needs to be started separately, outside the cache creation?

ferglor (approved):
Just some non blocking questions but lgtm!

ferglor (approved):
Groovy!

PR #277, opened 2023-09-15
Simulator log trigger events

EasterTheBunny:
This commit introduces some new functionality and some refactoring:

- new: log trigger events and log trigger upkeeps
- new: simulation progress tracking to the console
- change: RunBook changed to SimulationPlan
- change: SimulationPlan json altered for cleaner configurations
- change: loaders moved to loader package due to import cycles
- change: refactored main process for cleaner simulation exit

To run a new simulation:

```
make simulator && ./bin/simulator --simulate -f ./tools/simulator/plans/simplan_fast_check.json
```

Verification of expected performs is still not fully tested, but the updates above create the baseline for log trigger upkeeps.


cmd/simulator/config/event.go
```diff
@@ -0,0 +1,91 @@
...
+	ConditionalUpkeepType UpkeepType = "conditional"
+	LogTriggerUpkeepType  UpkeepType = "logTrigger"
+)
+
+// OCR3ConfigEvent is an event that indicates a new OCR config should be
+// broadcast. Consult libOCR for descriptions of each config var.
+type OCR3ConfigEvent struct {
+	Event
+	// F is the configurable faulty number of nodes
+	F int `json:"maxFaultyNodes"`
```
ferglor, line 31: Could we rename this to `MaxFaultyNodes` or `MaxFaultyNodesF`?

cmd/simulator/simulate/loader/upkeep.go
```diff
@@ -0,0 +1,72 @@
...
+		if !ok {
+			evts = []interface{}{}
+		}
+
+		create[upkeep.CreateInBlock.String()] = append(evts, chain.UpkeepCreatedTransaction{
+			Upkeep: upkeep,
+		})
+	}
+
+	if progress != nil {
```
ferglor, line 47: In what cases would progress be nil? Would it be worth even creating a no-op implementation we could default to to prevent nil checks like this?

simulation_runbooks/v3/runbook_fast_check.json
```diff
@@ -1,25 +1,31 @@
 {
-    "nodes": 4,
-    "maxNodeServiceWorkers": 100,
-    "maxNodeServiceQueueSize": 1000,
-    "avgNetworkLatency": "100ms",
-    "rpcDetail": {
+    "node": {
```
ferglor, line 2: Should we rename this directory from `simulation_runbooks` to `simulation_plans`?

cmd/simulator/config/event.go
EasterTheBunny: Changing this to be more detailed in the name helps describe it, but it also diverges from the libocr naming convention. They just call it `f`. I'm good either way. I can change it to be more in-line with the json config value for this config event though.

cmd/simulator/simulate/loader/upkeep.go
EasterTheBunny: That was just for testing since I didn't want to create a telemetry object for every unit test. I can probably come up with some sort of no-op to set the value in the constructor to remove later nil checks on that value.

simulation_runbooks/v3/runbook_fast_check.json
EasterTheBunny: Good catch.

cmd/simulator/config/event.go
ferglor: I suppose ideally the config value in the JSON would be called `f`, but I think it's better to have the struct keys aligned to the JSON config keys

cmd/simulator/simulate/loader/upkeep.go
ferglor: You could just create a mock with an empty `RegisterFn`?

amirylm:
We can clean the folders structure so everything will reside within the same package, e.g.
- `tools`
  - `simulator`
    - `config`
    - ...
    - `telemetry`
    - `runbooks` (or `plans`)

Then from `./cmd/simulator` just import everything you need for startup

cmd/simulator/config/keyring.go
```diff
@@ -54,14 +54,14 @@ func NewOffchainKeyring(encryptionMaterial, signingMaterial io.Reader) (*Offchai
 }
```
amirylm, line 54: Do we need to have a copy of the keyring or can we use the original one that is currently resides in `chainlink` repo? (and it was done just because of circular dependency)

cmd/simulator/simulate/loader/logtrigger.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// LogTriggerLoader ...
+type LogTriggerLoader struct {
+	// provided dependencies
+	progress ProgressTelemetry
+
+	// internal state values
+	mu       sync.RWMutex
+	triggers map[string][]interface{}
```
amirylm, line 17: Why interface rather than a concrete struct?

cmd/simulator/config/simulation.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if err := json.Unmarshal(encoded, &plan); err != nil {
+		return plan, fmt.Errorf("%w: failed to decode simulation plan: %s", ErrEncoding, err.Error())
+	}
+
+	plan.ConfigEvents = make([]OCR3ConfigEvent, 0)
+	plan.GenerateUpkeeps = make([]GenerateUpkeepEvent, 0)
+	plan.LogEvents = make([]LogTriggerEvent, 0)
+
+	type eventCollection struct {
```
amirylm, line 72: Can't we have a specific field for each event type, and avoid the "manual" decoding?

cmd/simulator/simulate/loader/logtrigger.go
EasterTheBunny: Probably just a leftover holding place. I'll make it a concrete type.

cmd/simulator/config/simulation.go
EasterTheBunny: The events are in a single JSON array and each event type has a `type` property for decoding. This JSON structure allows a person to visually order events by block number instead of having to jump between properties. All events can be viewed as if on a linear timeline.

```
"events": [
  ... event 1
  ... event 2
  ...
]
```

instead of the following each with their own ordering

```
"conditionalUpkeeps": [],
"logTriggerUpkeeps": [],
"ocrConfig": [],
"logTriggers": []
```

The intention was to make the config JSON more manageable with the consequence of adding more complexity to the unmarshaling.

amirylm: In that case, we can have a single struct for event, with pointer fields that can be nil depends on the actual type.

EasterTheBunny: I don't understand. Is your proposal to collapse all fields for all events into a single 'god object'? Right now, all events share a basic event structure:

```
type Event struct {
	Type         EventType `json:"type"`
	TriggerBlock *big.Int  `json:"eventBlockNumber"`
	Comment      string    `json:"comment,omitempty"`
}
```

And then each specific event extends the basic event like the following:

```
type SomeEvent struct {
	Event
	SomeSpecificProperty string `json:"someSpecificProperty"`
}
```

This method provides more strict types within Go with the consequence being a little more complexity in json decoding.

cmd/simulator/config/keyring.go
EasterTheBunny: I wanted to avoid pulling anything from `chainlink` because the dependency list grows a lot. The method for signing should not be important. All we need is to apply the interface.

cmd/simulator/config/simulation.go
amirylm: You can have a dedicated field for each extension, e.g.
```
type Event struct {
	Type         EventType `json:"type"`
	TriggerBlock *big.Int  `json:"eventBlockNumber"`
	Comment      string    `json:"comment,omitempty"`
	SomeEvent *SomeEvent `json:"someEvent,omitempty"`
}
```

This is a common pattern in golang because the lack of good inheritance (for example this is the way to go when you use protobuf).

EasterTheBunny: So then the JSON would look something like the following?

```
{
  "type": "theType",
  "eventBlockNumber": 10,
  "someEvent": {
    "someProp": 420
  }
}
```

How does this help? The simulation plan is already divided into distinct types and `Event` is never directly used. It only exists for decoding. There is a switch statement that in `tools/simulator/config/simulation.go` starting at line `88` that organizes the simulation events based on encoded type. This would still exist for your method to sort after decoding. The only difference would be a single `json.Unmarshal` instead of multiple for each event. Plus validation would need to exist such that a person should not be able to do the following:

```
{
  "type": "theType",
  "eventBlockNumber": 10,
  "someEvent": {
    "someProp": 420
  },
  "otherEvent": {
    "eventProp": "value"
  }
}
```

ferglor: I think I prefer having separate structs for each event, rather than having a single struct that acts as a catch-all for all events? I think it's just cleaner/more focused having a single struct per event type rather than  a single struct that you either have to have separate fields/objects per event OR if you try to "flatten" all the distinct event fields into the top level, it can get tricky trying to identify what type of event an instance of the god struct is representing, especially if multiple events have field names in common etc

amirylm: The Go-ish way is NOT to have such inheritance, the result is a messy & manual code for encoding/decoding.
You can find examples for this all over with protobufs, and also in our code we have `Trigger` which holds `LogTriggerExtension` which is exactly what I'm suggesting.

Protecting the user from doing mistakes can help in cases we have some external interface that might change over time, but in this case the users are internal (and technical).

We can go with this pattern JUST for the simulator, but this is something we should avoid doing in our operational code.

amirylm (dismissed):

amirylm (approved):

PR #280, opened 2023-09-22
Improve coordination of conditionals on successive performs

infiloop2:
1. In coordinated queue, if record already exists, allow adding a new proposal on a higher check block. This can happen in case of conditionals which perform very quickly and get coordinated on a new block after previous perform before expiry timeout

2. In coordinator, for conditonals, allow proposal to be processed on the same block as last perform. this matches the contract which allows check block to be equal last perform https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/automation/v2_1/KeeperRegistryBase2_1.sol#L771-L775 (only says stale report if `trigger.blockNum < transmitInfo.upkeep.lastPerformedBlockNumber`)

### DIFF
--- a/pkg/v3/stores/proposal_queue.go
+++ b/pkg/v3/stores/proposal_queue.go
@@ -48,7 +48,7 @@ func (pq *proposalQueue) Enqueue(newProposals ...ocr2keepers.CoordinatedBlockPro
 
 	for _, p := range newProposals {
 		if existing, ok := pq.records[p.WorkID]; ok {
-			if p.Trigger.BlockNumber <= existing.proposal.Trigger.BlockNumber {
+			if existing.proposal.Trigger.BlockNumber >= p.Trigger.BlockNumber {
 				// If existing proposal is on newer or equal check block then skip this proposal
 				continue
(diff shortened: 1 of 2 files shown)


RyanRHall (approved):
LGTM :+1:

pkg/v3/stores/proposal_queue.go
```diff
@@ -47,8 +47,11 @@ func (pq *proposalQueue) Enqueue(newProposals ...ocr2keepers.CoordinatedBlockPro
 	defer pq.lock.Unlock()
 
 	for _, p := range newProposals {
-		if _, ok := pq.records[p.WorkID]; ok {
-			continue
+		if existing, ok := pq.records[p.WorkID]; ok {
+			if existing.proposal.Trigger.BlockNumber >= p.Trigger.BlockNumber {
```
ferglor, line 51: Can any of these fields be nil?

infiloop2: good question! none of them are pointers so we should be good

PR #284, opened 2023-11-01
Retry Q: support custom intervals

amirylm:
AUTO-6958

The idea to have additional field for `CheckResult` to specify `RetryInterval`, enabling to aid-in custom intervals for retry records. In case no interval was specified, the default (30s) will be used.
This change will allow mercury lookup failures to have a shorter retry interval.

The changes are tested by adding a custom interval for records that are added to the Retry Q.
In addition, en/decoding tests were updated accordingly.


pkg/v3/stores/retry_queue.go
```diff
@@ -11,13 +11,17 @@ import (
 )
 
 var (
+	// DefaultExpiration is the default expiration time for items in the queue
 	DefaultExpiration = 24 * time.Hour
-	RetryInterval     = 30 * time.Second
+	// RetryInterval is the (default) minimum time between retries
```
FelixFan1992, line 16: it's not the minimum time anymore

pkg/v3/types/basetypes.go
```diff
@@ -14,6 +15,7 @@ const (
 var checkResultStringTemplate = `{
 	"PipelineExecutionState":%d,
 	"Retryable":%v,
+	"RetryInterval":%d,
```
FelixFan1992, line 18: is there a way to comment here to specify this is in seconds?

pkg/v3/postprocessors/retry.go
```diff
@@ -29,7 +29,10 @@ func (p *retryablePostProcessor) PostProcess(_ context.Context, results []ocr2ke
 	retryable := 0
 	for i, res := range results {
 		if res.PipelineExecutionState != 0 && res.Retryable {
```
RyanRHall, line 31: nitpick - can we change the 0 to the enum? Isn't there something like `ExecutionState.Success`?

pkg/v3/types/basetypes.go
RyanRHall: +1

pkg/v3/postprocessors/retry.go
amirylm: Good callout, but I'd prefer to do it in a separate PR to avoid extra noise.. created AUTO-7018

pkg/v3/types/basetypes.go
amirylm: We can comment above this string, but then maybe to explain other fields as well
P.S. I don't like this way of encoding (with string template) in any case, planning to remove it at some point.

FelixFan1992 (approved):

infiloop2 (approved):

PR #286, opened 2023-11-16
Optimize Sorting: shuffle string chars only once per workID

amirylm:
AUTO-7299

#### Description

As part of load testing, we found that our sorting doesn’t scale well as it takes too much CPU.
The motivation in this PR is to avoid shuffle the workID string multiple times, by doing it only once.

#### Changes

- `AddFromStagingHook` is using an additional slice to hold the shuffled strings, and sort that slice instead of the original one that requires to shuffle the workID in each sort operation (up to `n^2` operations)

#### Testing

There were no new tests added as the current cases cover this change and ensure we didn't break the function

### DIFF
--- a/pkg/v3/plugin/hooks/add_from_staging.go
+++ b/pkg/v3/plugin/hooks/add_from_staging.go
@@ -27,7 +27,7 @@ type AddFromStagingHook struct {
 }
 
 // RunHook adds results from the store to the observation.
-// It sorts by a shuffled workID. workID for all items is shuffled using a pseduorandom souce
+// It sorts by a shuffled workID. workID for all items is shuffled using a pseudorandom source
 // that is the same across all nodes for a given round. This ensures that all nodes try to
 // send the same subset of workIDs if they are available, while giving different priority
 // to workIDs in different rounds.
--- a/pkg/v3/plugin/hooks/add_from_staging.go
+++ b/pkg/v3/plugin/hooks/add_from_staging.go
@@ -46,7 +46,7 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
 		shuffledIDs[result.WorkID] = random.ShuffleString(result.WorkID, rSrc)
 	}
 	// sort by the shuffled workID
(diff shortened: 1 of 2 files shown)


pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -26,6 +26,11 @@ type AddFromStagingHook struct {
 	coord  types.Coordinator
 }
 
+// RunHook adds results from the store to the observation.
+// It sorts by a shuffled workID. workID for all items is shuffled using a pseduorandom souce
```
ferglor, line 30: *pseudorandom source

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -35,18 +40,18 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
...
-	// to workIDs in different rounds.
-	sort.Slice(results, func(i, j int) bool {
-		return random.ShuffleString(results[i].WorkID, rSrc) < random.ShuffleString(results[j].WorkID, rSrc)
+	// creating a map to hold the shuffled workIDs
+	shuffledIDs := make(map[string]string, len(results))
+	for _, result := range results {
+		shuffledIDs[result.WorkID] = random.ShuffleString(result.WorkID, rSrc)
+	}
+	// sort by the shuffled workID
+	sort.SliceStable(results, func(i, j int) bool {
```
ferglor, line 49: Is SliceStable needed here?

infiloop2 (commented):
nice find, lgtm! 
re. fergal's comment on Slice vs SliceStable, I think either should be fine since array has unique items but one might be faster

pkg/v3/plugin/hooks/add_from_staging.go
amirylm: The issue with Slice is that it is not deterministic in case of duplicated items.
I can revert as we don't expect to have workID duplications, and to have this change cross all other places in code to be consistent

ferglor (approved):

PR #290, opened 2023-11-24
Avoid encoding of RetryInterval

amirylm:
AUTO-7516

#### Motivation

Fixing a breaking change that was introduced when `RetryInterval` was added to `CheckResult` struct.

#### Description

Avoid the encoding and decoding of this property, by overriding MarshalJSON and UnmarshalJSON functions.

#### Changes

- added struct for encoding: checkResultMsg
- overriden Un/MarshalJSON functions
- aligned existing tests
- added new tests to ensure that the encoded object is not changed


ferglor (approved):
👍

pkg/v3/types/basetypes.go
```diff
@@ -216,12 +234,52 @@ func (r CheckResult) UniqueID() string {
...
+	}
+
+	return json.Marshal(crm)
+}
+
+func (r *CheckResult) UnmarshalJSON(data []byte) error {
+	var crm checkResultMsg
+
+	if err := json.Unmarshal(data, &crm); err != nil {
+		return err
```
infiloop2, line 265: can we add a temporary fix here which tries to unmarshall with an extra `RetryInterval` field for temporarily allowing compatibility with older versions

pkg/v3/types/basetypes.go
```diff
@@ -216,12 +234,52 @@ func (r CheckResult) UniqueID() string {
 
 func (r CheckResult) String() string {
 	return fmt.Sprintf(
-		checkResultStringTemplate, r.PipelineExecutionState, r.Retryable, r.RetryInterval, r.Eligible,
```
infiloop2, line 219: why remove from here?

pkg/v3/types/basetypes.go
infiloop2: thinking more, it will help in parsing observations from old versions, however it won't help in outcome consensus since that is done on raw bytes

pkg/v3/types/basetypes.go
amirylm: To avoid ambiguity, we use `String()` to encode into JSON, I don't want it to get into a report somehow

pkg/v3/types/basetypes.go
amirylm: In case older versions are running in parallel, they'll expect to have the `RetryInterval` so even if newer versions will read that property they will need to include it when producing outcome/report

pkg/v3/types/basetypes.go
infiloop2: hmm i see. still confusing for me personally but i guess any solution will have some level of confusion.

Maybe add a comment on top of this function as to how this should be modified? (e.g. we want to keep it consistent with marshallJSON?)

amirylm: Do we have a good reason why not to reduce that String function? i.e. to keep the standard golang printing of that object? If its just the hex encoding of PerformData we can use a custom type for it and reduce the function completely

infiloop2: yup agree this can be potentially cleaned up

infiloop2 (approved):

ferglor (approved):

PR #295, opened 2023-12-08
Verbose Simulation Output

EasterTheBunny:
By default, the simulation output was already verbose by printing logs for each node to separate files and printing simulation output to a file.

This change makes verbose no longer default and adds an option to do verbose logging output.

### DIFF
--- a/cmd/simulator/main.go
+++ b/cmd/simulator/main.go
@@ -26,6 +26,7 @@ import (
 var (
 	simulationFile  = flag.StringP("simulation-file", "f", "./simulation_plan.json", "file path to read simulation config from")
 	outputDirectory = flag.StringP("output-directory", "o", "./simulation_plan_logs", "directory path to output log files")
+	verbose         = flag.BoolP("verbose", "v", false, "make output verbose (prints logs to output directory)")
 	simulate        = flag.Bool("simulate", false, "run simulation")
 	serveCharts     = flag.Bool("charts", false, "create and serve charts")
 	profiler        = flag.Bool("pprof", false, "run pprof server on startup")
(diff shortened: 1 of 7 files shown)


cmd/simulator/main.go
```diff
@@ -26,6 +26,7 @@ import (
 var (
 	simulationFile  = flag.StringP("simulation-file", "f", "./simulation_plan.json", "file path to read simulation config from")
 	outputDirectory = flag.StringP("output-directory", "o", "./simulation_plan_logs", "directory path to output log files")
+	verbose         = flag.BoolP("verbose", "v", false, "make output verbose (prints logs to output directory)")
```
amirylm, line 29: What about having 2 flags here:
1. log level (info/debug/warn/error/none)
2. log output (<path to dir>, "." for outputDirectory or none for CLI)
  - i.e. to have the output directory as a base dir

EasterTheBunny: Adding level adds a good bit of scope because then one would need to decide what to log at each level. I think for now, to keep it simple, it should just be verbose (development logs) and not verbose (CI output).

The default directory is `./simulation_plan_logs` right now to make sure the output does not get included in git commits. This directory is included in `.gitignore` to help developers keep commits clean of any accidents. We could change it to the base directory, but then we need to update `.gitignore` as well.

amirylm: Our development logs are too informative so it might be hard to read, could have been better to have at least the level to let developers run with `info` or `warn` to get minimal output.

Regarding the the output, if we only enable to print logs to file system, it would be hard to understand issues in CI.
By printing to stdout we able to run anywhere while still being able to read the logs

PR #299, opened 2023-12-20
Move types into chainlink-common for LOOPP

ferglor:
The services created in core will need to come from the new AutomationProvider, as defined in chainlink-common

These services reference types from this repo; rather than make chainlink-common dependent on chainlink-automation, we need to instead move the automation types into the chainlink-common repo

## Associated PRs

chainlink: https://github.com/smartcontractkit/chainlink/pull/11631
chainlink-common: https://github.com/smartcontractkit/chainlink-common/pull/297


go.mod
```diff
@@ -11,14 +11,14 @@ require (
 	github.com/Maldris/mathparse v0.0.0-20170508133428-f0d009a7a773
 	github.com/ethereum/go-ethereum v1.12.0
 	github.com/go-echarts/go-echarts/v2 v2.2.6
-	github.com/google/uuid v1.3.0
+	github.com/google/uuid v1.3.1
 	github.com/jedib0t/go-pretty/v6 v6.4.7
 	github.com/pkg/errors v0.9.1
 	github.com/shopspring/decimal v1.3.1
-	github.com/smartcontractkit/libocr v0.0.0-20230922131214-122accb19ea6
+	github.com/smartcontractkit/libocr v0.0.0-20230925165524-ffa38fe11ef8
```
ferglor, line 18: libocr is being bumped here as a consequence of adding chainlink-common as a dependency. The only difference in this version of libocr and the previous version, is an update to the README.md.

pkg/v3/types/interfaces.go
```diff
@@ -1,109 +1,27 @@
...
-	Start(context.Context) error
-	Close() error
-}
-
-//go:generate mockery --name Ratio --structname MockRatio --srcpkg "github.com/smartcontractkit/chainlink-automation/pkg/v3/types" --case underscore --filename ratio.generated.go
-type Ratio interface {
-	// OfInt should return n out of x such that n/x ~ r (ratio)
-	OfInt(int) int
-}
+//go:generate mockery --name Ratio --structname MockRatio --srcpkg "github.com/smartcontractkit/chainlink-common/pkg/types/automation" --case underscore --filename ratio.generated.go
```
ferglor, line 27: We're moving the type definitions into common, but automation cares about the mocks, so we're keeping the mocks in this repo (by referencing the chainlink-common package in the generate command)

EasterTheBunny (approved):

cmalec (dismissed):

pkg/v3/plugin/ocr3_test.go
```diff
@@ -13,14 +13,15 @@ import (
...
 	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/smartcontractkit/libocr/offchainreporting2plus/ocr3types"
-	"github.com/smartcontractkit/libocr/offchainreporting2plus/types"
 	"github.com/stretchr/testify/assert"
 
 	ocr2keepers2 "github.com/smartcontractkit/chainlink-automation/pkg/v3"
 	"github.com/smartcontractkit/chainlink-automation/pkg/v3/plugin/hooks"
 	"github.com/smartcontractkit/chainlink-automation/pkg/v3/service"
-	ocr2keepers "github.com/smartcontractkit/chainlink-automation/pkg/v3/types"
+	types2 "github.com/smartcontractkit/chainlink-automation/pkg/v3/types"
```
amirylm, line 21: Let's use a better name than `types2`, maybe we can rename offchainreporting2plus/types to `ocr2plustypes` and keep this one as `types`

pkg/v3/types/interfaces.go
```diff
@@ -2,101 +2,77 @@ package types
 
 import (
 	"context"
+
+	"github.com/smartcontractkit/chainlink-common/pkg/types/automation"
 )
 
-type UpkeepTypeGetter func(UpkeepIdentifier) UpkeepType
-type WorkIDGenerator func(UpkeepIdentifier, Trigger) string
+//go:generate mockery --name Encoder --structname MockEncoder --srcpkg "github.com/smartcontractkit/chainlink-common/pkg/types/automation" --case underscore --filename encoder.generated.go
```
amirylm, line 9: Do we actually need to generate mocks here? can't we use generated mocks from cll-common?

tools/simulator/simulate/upkeep/perform_test.go
```diff
@@ -7,10 +7,13 @@ import (
 	"testing"
 	"time"
 
+	types2 "github.com/smartcontractkit/chainlink-automation/pkg/v3/types"
```
amirylm, line 10: rename types2

pkg/v3/types/interfaces.go
ferglor: Initially these mocks were migrated to common, but other developers weren't keen on having the mocks in common, as they weren't needed directly in common. So instead, the mocks are being generated in automation, since automation is the only repo that needs them

go.mod
infiloop2: is this the libocr version also used in chainlink-core?

infiloop2:
This change will make any automation type change a 3-repo dependency. While that might be needed, it would be good to have an overall design doc for what's the end vision for loops and what components will live where

go.mod
ferglor: No, core uses a newer version

amirylm (dismissed):

FelixFan1992 (approved):

PR #305, opened 2024-01-15
Dynamically calculate performables limit based on perform data size

amirylm:
AUTO-7597

#### Description

The idea is to put the maximal number of performables in observation/outcome dynamically calculate it based on performData and what fits into observation / outcome size, rather than using a fixed limit of number of upkeeps based on hardcoded limits.

#### Changes

* Added helper functions to calculate size
* Limit the amount of performables based on bytes size rather than hardcoded number of results
* Validate performables bytes size rather than hardcoded number of results
* Backwards compatibility (only for generating compatible outcome/observation, validation is being done only on the number of bytes)
  * **NOTE:** a followup PR should reduce backwards compatibility, which will be only for a single version

### DIFF
--- a/pkg/v3/plugin/performable.go
+++ b/pkg/v3/plugin/performable.go
@@ -15,11 +15,11 @@ type resultAndCount struct {
 }
 
 type performables struct {
-	limit           int
-	keyRandSource   [16]byte
-	quorumThreshold int
-	logger          *log.Logger
-	resultCount     map[string]resultAndCount
+	limit, sizeLimit int
+	keyRandSource    [16]byte
+	quorumThreshold  int
+	logger           *log.Logger
+	resultCount      map[string]resultAndCount
 }
 
 // Performables gets quorum on agreed check results which should ultimately be
@@ -27,10 +27,11 @@ type performables struct {
 // and simply adds all results which achieve the quorumThreshold.
 // Results are agreed upon by their UniqueID() which contains all the data
 // within the result.
-func newPerformables(quorumThreshold int, limit int, rSrc [16]byte, logger *log.Logger) *performables {
+func newPerformables(quorumThreshold int, limit, sizeLimit int, rSrc [16]byte, logger *log.Logger) *performables {
 	return &performables{
 		quorumThreshold: quorumThreshold,
 		limit:           limit,
+		sizeLimit:       sizeLimit,
 		keyRandSource:   rSrc,
 		logger:          logger,
 		resultCount:     make(map[string]resultAndCount),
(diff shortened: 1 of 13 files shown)


EasterTheBunny:
What if we estimated size in the check pipeline instead of in the plugin? In this way, size just becomes a number and the size estimation logic is closer to the encoding functions. It would make the process less brittle.

pkg/v3/types/basetypes.go
```diff
@@ -163,6 +163,21 @@ type CheckResult struct {
 	RetryInterval time.Duration
 }
 
+// Size returns the size of the check result in bytes
+func (cr *CheckResult) Size() int {
```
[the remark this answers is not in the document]
amirylm: Note that all the numbers have a corresponding comment to specify what field they represent.
The most naive option is to encode the object and count the bytes, but that would be expensive in runtime (encoding/decoding is currently taking dozens percentage of CPU)

The CheckResult currently still resides here, once moved to cl-common I'll adjust.

amirylm:
> What if we estimated size in the check pipeline instead of in the plugin? In this way, size just becomes a number and the size estimation logic is closer to the encoding functions. It would make the process less brittle.

This change refers to Observation and Outcome which are encoded within the plugin as they are used by OCR (no EVM encoding). 
One thing that could be an issue is the overhead of JSON encoding (property names, {}, quotes, white spaces)

EasterTheBunny:
Ahh ok. That makes sense. This is for byte limits of Observation and Outcome.

pkg/v3/types/basetypes.go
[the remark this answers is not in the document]
infiloop2: +1 don't quite like this hardcoded size counting as historically this as bitten us in the past. 

1. Anytime the struct is changed it creates an implicit dependency to change which is very easy to forget
2. If encoding type is changed then this needs to change
3. This will likely need some magic numbers to match exact json encoding size then just adding the byte size of fields

pkg/v3/plugin/performable.go
```diff
@@ -82,10 +83,25 @@ func (p *performables) set(outcome *ocr2keepersv3.AutomationOutcome) {
 		return random.ShuffleString(performable[i].WorkID, p.keyRandSource) < random.ShuffleString(performable[j].WorkID, p.keyRandSource)
 	})
 
+	// TODO: remove this in next version, it's a temporary fix for
+	// supporting old nodes that will limit the number of results rather than the size
+	// of the outcome
```
infiloop2, line 88: Need to think more deeply, but i guess this will be a breaking change even if we remove this in the next version? 

I think the ideal way to make this non breaking is to do it with a feature flag that is then turned on at once for all nodes (via config), but I think for this specific change we may just be ok to just take a breaking risk as we don't expect more than 50 performables in a round for current traffic

amirylm: > Need to think more deeply, but i guess this will be a breaking change even if we remove this in the next version?

In case all ops will update to this version, where the nodes will produce "limited" Outcome/Observation but won't validate that, they can update to next version where we'll start producing "unlimited" Outcome/Observation as the validation will accept those

> for this specific change we may just be ok to just take a breaking risk as we don't expect more than 50 performables in a round for current traffic

OK make sense

infiloop2: > they can update to next version where we'll start producing "unlimited" Outcome/Observation as the validation will accept those

Yes the validation would pass but outcome consensus would still not happen if some NOPs produce limited outcome and some with more

PR #314, opened 2024-02-23
Update the OffchainConfig with a LogProviderConfig struct

ferglor:
https://smartcontract-it.atlassian.net/browse/AUTO-9023

In this PR, we're updating the OffchainConfig with a LogProviderConfig struct. This new struct holds configuration values for the log event provider.

# Related PRs

- chainlink: https://github.com/smartcontractkit/chainlink/pull/12152
- chainlink-common: https://github.com/smartcontractkit/chainlink-common/pull/387


pkg/v3/config/config.go
```diff
@@ -63,6 +63,17 @@ type OffchainConfig struct {
...
+	// LogProviderConfig holds configuration for the log provider
+	LogProviderConfig LogProviderConfig `json:"logProviderConfig"`
+}
+
+type LogProviderConfig struct {
+	// NumOfLogUpkeeps is the number of log upkeeps supported by the registry
+	NumOfLogUpkeeps uint32 `json:"numOfLogUpkeeps"`
+
+	// FastExecLogsHigh is the maximum number of logs that we are committed to processing for each upkeep
+	FastExecLogsHigh uint32 `json:"fastExecLogsHigh"`
```
amirylm, line 76: missing `FastExecLogsLow`

amirylm (dismissed):

ogtownsend (approved):

pkg/v3/config/config.go
```diff
@@ -63,6 +63,20 @@ type OffchainConfig struct {
...
```
infiloop2, line 76: where do we define the default values for these?

pkg/v3/config/config.go
```diff
@@ -63,6 +63,20 @@ type OffchainConfig struct {
...
+
+	// FastExecLogsLow is the minimum number of logs that we are committed to processing for each upkeep
+	FastExecLogsLow uint32 `json:"fastExecLogsLow"`
```
infiloop2, line 79: I see this is not actually being passed to log provider. Is that expected?

pkg/v3/config/config.go
```diff
@@ -63,6 +63,20 @@ type OffchainConfig struct {
 
 	// MaxUpkeepBatchSize is the max upkeep batch size of the OCR2 report.
 	MaxUpkeepBatchSize int `json:"maxUpkeepBatchSize"`
+
+	// LogProviderConfig holds configuration for the log provider
+	LogProviderConfig LogProviderConfig `json:"logProviderConfig"`
+}
+
+type LogProviderConfig struct {
+	// NumOfLogUpkeeps is the number of log upkeeps supported by the registry
```
infiloop2, line 72: can we expand on this. `NumOfLogUpkeeps` is the number of upkeeps for which we give `FastExecLogsHigh` and `FastExecLogsLow` guarantees. If there are more upkeeps registered then upkeeps will still be performed but the guarantees may get violated

infiloop2: cc @amirylm @ferglor please confirm if the above expectation is correct

pkg/v3/config/config.go
infiloop2: Can we define these concretely. Are these per block / per second. Do we need float values

infiloop2 (changes requested):
Requesting changes for clearer definition of added configs

pkg/v3/config/config.go
infiloop2: Depending on this we might need to add a block time config.

pkg/v3/config/config.go
ferglor: Yes, we aren't using this field yet but it was going to be added eventually so I added it now, since all three fields have been identified as part of the scaling design

infiloop2: ok, would still be good to pass to factory.logProvider.SetConfig(commontypes.LogEventProviderConfig{
 		NumLogUpkeeps:    conf.LogProviderConfig.NumOfLogUpkeeps,
 		FastExecLogsHigh: conf.LogProviderConfig.FastExecLogsHigh,
 	}) below so that core has access to the config and can use it when it is ready

pkg/v3/config/config.go
amirylm: The idea was that these are per block, so yes we need to use floats.
Block time is not needed as we can calculate it.

pkg/v3/config/config.go
amirylm: Will be handled in following PRs

PR #317, opened 2024-04-01
Scale observation building with massive amount of results

amirylm:
AUTO-9416

#### Description

Currently we sort the entire set of staged upkeeps by random seed order based on the OCR round/seq number, then take a limit of items to be added to observation. 
As part of the sorting we are doing string shuffling with the random seed, which is CPU+memory intensive work that makes the entire operation slow, and breaks when scaling as we repeat the same expensive operation to get random order in each round. Eventually resulting in multiple “Observation is taking too long” warnings or even crushes the nodes.

We will use a random number every x rounds (e.g. 10) and maintain a cache for the shuffled work ids.
A map will be used to store shuffled ids, it will get cleaned every time the random source changes (10 rounds).

#### Changes

- Added staging hook state with cache and last src
- Populate cache before sorting
- Use cache when sorting

### DIFF
--- a/pkg/v3/plugin/hooks/add_from_staging.go
+++ b/pkg/v3/plugin/hooks/add_from_staging.go
@@ -49,8 +49,7 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
 	}
 
 	results = hook.sorter.orderResults(results, rSrc)
-	n := len(results)
-	if n > limit {
+	if n := len(results); n > limit {
 		results = results[:limit]
 		hook.logger.Printf("skipped %d available results in staging", n-limit)
 	}


infiloop2 (commented):
PR looks great to me!

Could we add some unit test coverage for the updated logic

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -39,20 +53,44 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
...
 		return err
 	}
-	// creating a map to hold the shuffled workIDs
-	shuffledIDs := make(map[string]string, len(results))
-	for _, result := range results {
-		shuffledIDs[result.WorkID] = random.ShuffleString(result.WorkID, rSrc)
+
+	results = hook.orderResults(results, rSrc)
+	n := len(results)
+	if n > limit {
```
ferglor, line 59: `n` isn't used outside the if so we can collapse line 58 and line 59 into `if n := len(results); n > limit`

ferglor (dismissed):

PR #321, opened 2024-04-09
Switch from encoding/json to goccy/go-json

ferglor:
https://smartcontract-it.atlassian.net/browse/AUTO-9387

In this PR, we're switching from `encoding/json` to `goccy/go-json`, a library with [promising benchmarks](https://github.com/goccy/go-json?tab=readme-ov-file#benchmarks) that is `encoding/json` compatible

In terms of acceptance criteria for this change:
- We should expect all tests to pass (node upgrade test too)
- We will hopefully see performance improvements (fewer "Outcome/Reports is taking too long" messages during load tests)

One point worth considering is that the error messages returned by `goccy/go-json` are more specific than those returned by `encoding/json`; as such, we should review the automation code in the plugin and in core, for any instances where we check for a specific error message, and ensure that the same flow is executed when we use `goccy/go-json`

# Testing

For this test, the load test was ran against this branch, and latest develop, multiple times. In each set of test runs, the shape of the log volumes reported in grafana was the same.

## "Taking too long" logs

Based on the logs we see in grafana, and given the same load test scenario (500 upkeeps over one hour), it looks as though there has been a decrease in the number of "taking too long" logs overall.

### Before 

image

### After 

image

## "Observation is taking too long" logs

Observations taking too long still happens, with only a slight reduction in occurrences of long observations

### Before

image

### After 

image

## "Outcome is taking too long" logs

This is the most significant improvement we've gained with this change (note the Y axis)

### Before

image

### After 

image

## Integration tests

Integration tests are passing for this change as per this draft PR: https://github.com/smartcontractkit/chainlink/pull/12755

## Node upgrade tests

Node upgrade tests are passing for this change as per this CI run: https://github.com/smartcontractkit/chainlink/actions/runs/8880827638

## Summary

It looks as though this change does give us some benefit, especially for outcomes, though it looks like there are other causes of observations taking too long, unrelated to JSON encoding/decoding

### DIFF
--- a/pkg/v3/plugin/ocr3_test.go
+++ b/pkg/v3/plugin/ocr3_test.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"io"
 	"log"
@@ -13,8 +12,7 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/crypto"
-	"github.com/stretchr/testify/assert"
-
+	"github.com/goccy/go-json"
 	ocr2keepers2 "github.com/smartcontractkit/chainlink-automation/pkg/v3"
 	"github.com/smartcontractkit/chainlink-automation/pkg/v3/plugin/hooks"
 	"github.com/smartcontractkit/chainlink-automation/pkg/v3/service"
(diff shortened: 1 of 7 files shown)


pkg/v3/plugin/ocr3_test.go
```diff
@@ -942,7 +941,7 @@ func TestOcr3Plugin_ValidateObservation(t *testing.T) {
 			name:        "validating an empty observation returns an error",
 			observation: ocr2plustypes.AttributedObservation{},
 			expectsErr:  true,
-			wantErr:     errors.New("unexpected end of JSON input"),
```
amirylm, line 945: Have you checked if we rely on some of these errors in the core node?

ferglor: From what I can see, the code paths in the plugin that are affected by this import are called by libocr, and I haven't seen any error message checking in libocr; instead it just checks for non nil errors and uses those to return a true/false response to the caller (also in libocr)

amirylm (approved):

PR #322, opened 2024-04-10
Reduce sorting impact on CPU for staged results

amirylm:
#### Description

Sorting the entire result set becomes to heavy in case there is a massive amount of staged results.
The idea is to sort only first items instead of sorting the whole result set.

NOTE: this is a leftover of #317 

#### Changes

- Added (limit int) in the order func
- Terminate sort after we have the first (limit) elements sorted

#### Test Plan

- Unit tests have been added


infiloop2 (approved):

ferglor (approved):

amirylm:
Will be fixed in AUTO-10123

PR #325, opened 2024-05-03
Dynamically add performables to the observation until MaxObservationLength is exceeded

ferglor:
https://smartcontract-it.atlassian.net/browse/AUTO-9371

In this PR, we're changing how performables (`CheckResults`) are added to the observation. Rather than always adding a fixed number of performables to the observation, we're now adding performables to the observation, while the observation still has capacity for more performables without exceeding the maximum observation length

## The changes

- Implement a number of `Length` functions for various types that are used within the Observation type
  - These functions calculate the size in bytes of raw JSON representation for specific values 
- Implement three functions for adding performables to the observation (only one will be merged, all three have been included for illustrative purposes)
  - `AddByJSON` - JSON encodes the observation every time we add a performable (accurate, but very expensive)
  - `AddByEstimate` - Use the new estimates package to calculate the raw JSON size of the observation every time we add a performable (inaccurate, but very cheap)
  - `AddByEstimateAggressive` - Uses a combination of the above two functions (accurate, still very cheap)

## Benchmarks

- `AddByJSON`:   ~1315313291 ns/op 🥉 
- `AddByEstimate`:   ~768741 ns/op 🥇 
- `AddByEstimateAggressive`:  ~876293 ns/op 🥈 

## Questions 

- In testing, observations with hundreds of performables have been observed. Should we still apply some limit to the number of performable, e.g. 100?
  - If yes, should we stick to the simpler but less efficient implementation of `AddByEstimate` for adding performables to the observation?

### DIFF
--- a/pkg/v3/observation_test.go
+++ b/pkg/v3/observation_test.go
@@ -577,7 +577,7 @@ func TestLargeObservationSize(t *testing.T) {
 		})
 	}
 	largePerformData := [10001]byte{}
-	for i := 0; i < ObservationPerformablesLimit; i++ {
+	for i := 0; i < 67; i++ {
 		newResult := validLogResult
 		uid := commontypes.UpkeepIdentifier{}
 		uid.FromBigInt(big.NewInt(int64(i + 10001)))
--- a/pkg/v3/observation_test.go
+++ b/pkg/v3/observation_test.go
@@ -577,7 +577,7 @@ func TestLargeObservationSize(t *testing.T) {
 		})
 	}
 	largePerformData := [10001]byte{}
-	for i := 0; i < 67; i++ {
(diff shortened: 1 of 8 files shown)


pkg/v3/observation_test.go
```diff
@@ -577,7 +577,7 @@ func TestLargeObservationSize(t *testing.T) {
 		})
 	}
 	largePerformData := [10001]byte{}
-	for i := 0; i < ObservationPerformablesLimit; i++ {
```
ferglor, line 580: I'm not sure what to do here; should this test still add the maximum number of performables (which will always exceed the max observation length), or should we change the test to expect the max observation length to be exceeded?

amirylm: Expect the max observation length to be exceeded, the number of elements is dynamic anyway

PR #328, opened 2024-05-14
add custom golangci-lint file; fix containedctx issues

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3203

This PR adds a custom linter config file with the standard settings, and cleans up all cases of the `containedctx` rule.
> containedctx is a linter that detects struct contained context.Context field. This is discouraged technique in favour of passing context as first argument of method or function. For rationale please read [Contexts and structs](https://go.dev/blog/context-and-structs) the Go blog post.


.github/workflows/go.yml
```diff
@@ -22,9 +22,9 @@ jobs:
       uses: golangci/golangci-lint-action@9d1e0624a798bb64f6c3cea93db47765312263dc # v5.1.0
       with:
         # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-        version: v1.55.2
+        version: v1.58.1
         # only-new-issues is only applicable to PRs, otherwise it is always set to false
-        only-new-issues: false # disabled for PRs due to unreliability
+        only-new-issues: true
```
jmank88, line 27: There is too much to fix, so I think we need to flip this back and hope for the best :crossed_fingers:

PR #330, opened 2024-05-30
Add up to 100 performables based on percentage exceeded

ferglor:
See these two previous PRs for context on this change: 

- https://github.com/smartcontractkit/chainlink-automation/pull/325  
- https://github.com/smartcontractkit/chainlink-automation/pull/326 

This change estimates the average size of the performables that have been added, and when we exceed the max observation length, we calculate how far over the max length we are, as a percentage of the total length of performables, and then reduce the number of performables to add based on the average performable size vs the number of bytes we exceeded the max length by. 

When dealing with heavily populated performables (i.e. large perform data), we can fit in ~66 performables into the observation

When dealing with lightly populated performables, we can fit in almost 600 performables into the observation

With this change however, we are applying a limit of 100 performables in the observation, though this can be easily bumped

Node upgrade tests passing: https://github.com/smartcontractkit/chainlink/actions/runs/9318732991
Passing load test (verified through slack): https://github.com/smartcontractkit/chainlink/actions/runs/9318760597

### DIFF
--- a/pkg/v3/observation.go
+++ b/pkg/v3/observation.go
@@ -13,7 +13,7 @@ import (
 // as different nodes would upgrade at different times and would need to
 // adhere to each others' limits
 const (
-	ObservationPerformablesLimit          = 50
+	ObservationPerformablesLimit          = 100
 	ObservationLogRecoveryProposalsLimit  = 5
 	ObservationConditionalsProposalsLimit = 5
 	ObservationBlockHistoryLimit          = 256
--- a/pkg/v3/observation_test.go
+++ b/pkg/v3/observation_test.go
@@ -609,7 +609,7 @@ func TestLargeObservationSize(t *testing.T) {
 	assert.NoError(t, err, "no error in decoding valid automation observation")
 
 	assert.Equal(t, ao, decoded, "final result from encoding and decoding should match")
-	assert.Less(t, len(encoded), MaxObservationLength, "encoded observation should be less than maxObservationSize")
+	assert.Greater(t, len(encoded), MaxObservationLength, "encoded observation will exceed maxObservationSize")
 }
 
 func mockUpkeepTypeGetter(id commontypes.UpkeepIdentifier) types.UpkeepType {
--- a/pkg/v3/observation_test.go
+++ b/pkg/v3/observation_test.go
@@ -576,16 +576,6 @@ func TestObservationSizeWithEmptyPeformables(t *testing.T) {
 			Hash:   [32]byte{1},
 		})
 	}
-	//performData := [5001]byte{}
-	//for i := 0; i < ObservationPerformablesLimit; i++ {
-	//	newResult := validLogResult
-	//	uid := commontypes.UpkeepIdentifier{}
-	//	uid.FromBigInt(big.NewInt(int64(i + 10001)))
-	//	newResult.UpkeepID = uid
-	//	newResult.WorkID = mockWorkIDGenerator(newResult.UpkeepID, newResult.Trigger)
-	//	newResult.PerformData = performData[:]
-	//	ao.Performable = append(ao.Performable, newResult)
-	//}
 	for i := 0; i < ObservationConditionalsProposalsLimit; i++ {
 		newProposal := validConditionalProposal
 		uid := commontypes.UpkeepIdentifier{}
--- a/pkg/v3/observation_test.go
+++ b/pkg/v3/observation_test.go
@@ -564,7 +564,7 @@ func TestInvalidLogProposal(t *testing.T) {
 	assert.ErrorContains(t, err, "log trigger extension cannot be empty for log upkeep")
 }
 
-func TestLargeObservationSize(t *testing.T) {
+func TestObservationSizeWithEmptyPeformables(t *testing.T) {
 	ao := AutomationObservation{
 		Performable:     []commontypes.CheckResult{},
 		UpkeepProposals: []commontypes.CoordinatedBlockProposal{},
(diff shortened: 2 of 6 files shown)


pkg/v3/observation.go
```diff
@@ -13,7 +13,7 @@ import (
 // as different nodes would upgrade at different times and would need to
 // adhere to each others' limits
 const (
-	ObservationPerformablesLimit          = 50
+	ObservationPerformablesLimit          = 100
```
infiloop2, line 16: 🎉

pkg/v3/observation_test.go
```diff
@@ -609,7 +609,7 @@ func TestLargeObservationSize(t *testing.T) {
 	assert.NoError(t, err, "no error in decoding valid automation observation")
 
 	assert.Equal(t, ao, decoded, "final result from encoding and decoding should match")
-	assert.Less(t, len(encoded), MaxObservationLength, "encoded observation should be less than maxObservationSize")
+	assert.Greater(t, len(encoded), MaxObservationLength, "encoded observation will exceed maxObservationSize")
```
infiloop2, line 612: this test doesn't make much sense now. maybe we can remove it all together if we have a test somewhere else that verifies that observation size will not exceed MaxObservationLength in any case

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -49,16 +50,40 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
 	}
 
 	results = hook.sorter.orderResults(results, rSrc)
-	if n := len(results); n > limit {
-		results = results[:limit]
-		hook.logger.Printf("skipped %d available results in staging", n-limit)
-	}
-	hook.logger.Printf("adding %d results to observation", len(results))
-	obs.Performable = append(obs.Performable, results...)
+	added, _ := hook.addByPercentageExceeded(obs, limit, results, ocr2keepersv3.ObservationPerformablesLimit)
```
infiloop2, line 53: not sure how `ocr2keepersv3.ObservationPerformablesLimit` makes sense here as base size? It will be always 100 while I understand base size is the current size in bytes?

pkg/v3/plugin/ocr3.go
```diff
@@ -74,13 +81,6 @@ func (plugin *ocr3Plugin) Observation(ctx context.Context, outctx ocr3types.Outc
 	}
 	prommetrics.AutomationPluginPerformables.WithLabelValues(prommetrics.PluginStepResultStore).Set(float64(len(observation.Performable)))
 
-	if err := plugin.AddLogProposalsHook.RunHook(&observation, ocr2keepersv3.ObservationLogRecoveryProposalsLimit, getRandomKeySource(plugin.ConfigDigest, outctx.SeqNr)); err != nil {
```
infiloop2, line 77: nit. can we add a comment somewhere that AddFromStagingHook should always be the last hook that is called as it ensures the size constraints are met

pkg/v3/observation_test.go
infiloop2: or maybe we can keep this test to verify that with max proposals, we still have non zero space left for performables which are filled later

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -49,16 +50,40 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
...
+
+	encodings := 1
+	b, _ := obs.Encode()
+
+	if observationSize := len(b); observationSize > ocr2keepersv3.MaxObservationLength {
+		performablesSize := observationSize - baseSize
+		avgPerformableSize := performablesSize / limit
+		exceededBy := observationSize - ocr2keepersv3.MaxObservationLength
+		avgPerformablesExceeded := int(math.Ceil(float64(exceededBy) / float64(avgPerformableSize)))
+		limit -= avgPerformablesExceeded + 1 // ensure we always remove at least one performable on the next call
```
infiloop2, line 78: can we check here that limit does not go below 0?

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -49,16 +50,40 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
...
 	return nil
 }
 
+func (hook *AddFromStagingHook) addByPercentageExceeded(obs *ocr2keepersv3.AutomationObservation, limit int, results []automation.CheckResult, baseSize int) (int, int) {
+	if limit > len(results) {
+		limit = len(results)
+	}
+	existingPerformables := obs.Performable
+
+	obs.Performable = append(obs.Performable, results[:limit]...)
```
infiloop2, line 68: i don't fully follow the logic of why we are appending here instead of just setting the observables. 

It would work directly if we were setting 
`obs.Performable = results[:limit]` ?

pkg/v3/plugin/hooks/add_from_staging.go
```diff
@@ -49,16 +50,40 @@ func (hook *AddFromStagingHook) RunHook(obs *ocr2keepersv3.AutomationObservation
...
+
+	encodings := 1
+	b, _ := obs.Encode()
+
+	if observationSize := len(b); observationSize > ocr2keepersv3.MaxObservationLength {
+		performablesSize := observationSize - baseSize
+		avgPerformableSize := performablesSize / limit
```
infiloop2, line 75: can limit reach 0 here?

pkg/v3/plugin/ocr3_test.go
```diff
@@ -635,12 +635,14 @@ func TestOcr3Plugin_Observation(t *testing.T) {
...
-			AddBlockHistoryHook:    hooks.NewAddBlockHistoryHook(metadataStore, logger),
-			AddFromStagingHook:     hooks.NewAddFromStagingHook(resultStore, coordinator, logger),
-			Logger:                 logger,
+			RemoveFromStagingHook:       hooks.NewRemoveFromStagingHook(resultStore, logger),
+			RemoveFromMetadataHook:      hooks.NewRemoveFromMetadataHook(metadataStore, logger),
+			AddToProposalQHook:          hooks.NewAddToProposalQHook(proposalQueue, logger),
+			AddBlockHistoryHook:         hooks.NewAddBlockHistoryHook(metadataStore, logger),
+			AddFromStagingHook:          hooks.NewAddFromStagingHook(resultStore, coordinator, logger),
+			AddLogProposalsHook:         hooks.NewAddLogProposalsHook(metadataStore, coordinator, logger),
+			AddConditionalProposalsHook: hooks.NewAddConditionalProposalsHook(metadataStore, coordinator, logger),
```
infiloop2, line 644: not sure what changed, do we know why we need to pass these now? were they not needed before?

pkg/v3/plugin/hooks/add_from_staging.go
ferglor: The previous functionality appended performables, and we had tests that expected the append vs a set so I kept it this way. Happy to change this though!

pkg/v3/plugin/ocr3_test.go
ferglor: It's because we've changed the ordering of when we add performables; now these two hooks get executed first, and without this change some tests hit a nil pointer because these two hooks had not been set

pkg/v3/plugin/hooks/add_from_staging.go
ferglor: It doesn't actually, this is a hangover from other testing I was doing, nice catch!!

pkg/v3/plugin/hooks/add_from_staging.go
ferglor: No but I can add a check just in case, or at least a comment

amirylm (approved):
LGTM

infiloop2 (approved):

PR #335, opened 2024-10-13
audit contexts

jmank88:
https://smartcontract-it.atlassian.net/browse/BCF-3204

Supports:
- https://github.com/smartcontractkit/chainlink/pull/14920


pkg/v3/flows/conditional_test.go
jmank88, line 1: The mock constructors automatically register the `*testing.T` and `AssertExpectations` on `t.Cleanup`.

pkg/v3/types/mocks/metadatastore.generated.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.28.1. DO NOT EDIT.
+// Code generated by mockery v2.43.2. DO NOT EDIT.
```
jmank88, line 1: I didn't mean to use a different version here. Shouldn't CI be failing?

jmank88: `.tool-versions` was on 2.12.1 even though these files were on 2.28.1

jmank88: Bumped to 2.43.2

pkg/v3/tickers/time.go
```diff
@@ -38,13 +40,13 @@ func NewTimeTicker[T any](interval time.Duration, observer observer[T], getterFn
...
-
-	if !t.closer.Store(cancel) {
-		return fmt.Errorf("already running")
+func (t *timeTicker[T]) Start(ctx context.Context) error {
+	if err := t.StartOnce("timeTicker", func() error { return nil }); err != nil {
+		return err
 	}
+	defer close(t.done)
+	ctx, cancel := t.stopCh.Ctx(ctx)
+	defer cancel()
```
infiloop2, line 49: can't comment on the exact lines, 
in lines 64-66 (in the for loop) should there be a similar statement as above file 
```
if ctx.Err() != nil {
 					return nil
 				}
```

infiloop2 (commented):
looks good, added a small comment

pkg/v3/tickers/time.go
jmank88: Updated to `continue`

infiloop2: yeah that does seem like a bug. thanks for fixing!

infiloop2 (dismissed):

pkg/v3/tickers/time.go
jmank88: Had to fix a test too

infiloop2 (approved):

PR #339, opened 2026-04-01
reduce log message length

jmank88:
Including full results produces large log lines, up to 170KiB.

### DIFF
--- a/pkg/v3/observer.go
+++ b/pkg/v3/observer.go
@@ -111,7 +111,10 @@ func (o *Observer[T]) Process(ctx context.Context, tick tickers.Tick[[]T]) error
 		return err
 	}
 
-	o.lggr.Printf("finished processing of %d results: %+v", len(results), results)
+	o.lggr.Printf("finished processing of %d results", len(results))
+	for i, result := range results {
+		o.lggr.Printf("result %d: %+v", i+1, result)
+	}
 
 	return nil
 }


pkg/v3/observer.go
```diff
@@ -111,7 +111,7 @@ func (o *Observer[T]) Process(ctx context.Context, tick tickers.Tick[[]T]) error
 		return err
```
patrickhuie19, line 111: /nit logging this err may be helpful

patrickhuie19 (dismissed):

patrickhuie19:
cc @ibrajer 

2 concerns here:

1. is there any useful signal in this log line?
2. given the length (170KiB), is there a more clever way to extract the signal that is needed for debugging?

infiloop2 (approved):

infiloop2 (changes requested):
actually i am looking through the runbooks and this log line is useful to debug the result of evaluation of an upkeep (whether it was eligible/ineliglible and for what reason) https://github.com/smartcontractkit/ir/blob/main/runbooks/Alert_Runbooks/automation_runbooks.md?plain=1#L539-L547

So at the very least we should extract this information from the results and log them. 

My preference though would be to leave this untouched for now as we might need the other information too to debug an edge case in future. Are NOPs complaining right now because of this log line?

jmank88:
> actually i am looking through the runbooks and this log line is useful to debug the result of evaluation of an upkeep (whether it was eligible/ineliglible and for what reason) [smartcontractkit/ir@`081e1dc`/runbooks/Alert_Runbooks/automation_runbooks.md?plain=1#L539-L547](https://github.com/smartcontractkit/ir/blob/081e1dc2bc3e56faa51a5056d2b9ce933a9757f4/runbooks/Alert_Runbooks/automation_runbooks.md?plain=1#L539-L547)
> 
> So at the very least we should extract this information from the results and log them.
> 
> My preference though would be to leave this untouched for now as we might need the other information too to debug an edge case in future. Are NOPs complaining right now because of this log line?

Infra is already truncating logs at 1MB, and wants to eventually reduce that to just 16KB.

infiloop2:
>Infra is already truncating logs at 1MB, and wants to eventually reduce that to just 16KB.

Maybe for now we can log each result array member in a separate line? This doesn't reduce overall cost however maintains our debuggability without investing more time into alternate solutions. We aim to deprecate automation overall when cre is ready and recover the cost then

jmank88:
> Maybe for now we can log each result array member in a separate line? This doesn't reduce overall cost however maintains our debuggability without investing more time into alternate solutions. We aim to deprecate automation overall when cre is ready and recover the cost then

Updated

anirudhwarrier (approved):

